PR #199, opened 2020-09-15
[WIP] [LDR R2] EOS-12351 S3 - Kafka support for S3 background delete

t7ko-seagate:
* EOS-12351:S3:Kafka support for S3 background delete

Co-authored-by: Rajesh Nambiar <rajesh.nambiar@seagate.com>
Signed-off-by: Ivan Tishchenko <ivan.tishchenko@seagate.com>

### DIFF
--- a/s3backgrounddelete/kafka_setup.md
+++ b/s3backgrounddelete/kafka_setup.md
@@ -1,0 +1,34 @@
+### License
+
+Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


s3backgrounddelete/kafka_setup.md
```diff
@@ -0,0 +1,34 @@
+### License
+
```
ujjwalpl, line 2: All the kafka related code should be checked into cortx-utils and it should be independent of S3. There should be clearly carved out interfaces to CORTX Message bus where S3 background delete can push or subscribe for the messages. Interfaces needs to be generic and independent of S3.

ujjwalpl (changes requested):

PR #249, opened 2020-09-28
EOS-13729 Add proper logs in case ldap and auth password is not in sync.

abhilekh268:


auth/server/src/main/java/com/seagates3/dao/ldap/LdapConnectionManager.java
```diff
@@ -69,6 +69,7 @@ public static void initLdap()
             String msg = "Failed to initialise LDAP.\n" + ex.toString();
             if (retryLdapConnection(ex.getResultCode())) {
               LOGGER.error(msg);
+              LOGGER.debug(msg);
```
amitwac2608, line 72: Why we need debug log when error is already here ? 

Are we missing something ? Error log level has more severity then debug, so error logging is enough.

abhilekh268:
not needed.we are already loging error

PR #250, opened 2020-09-29
Removed empty 'Group' tag from scripttest spec file.

amitwac2608:
Because of the empty tag, rpm creation was failing.


shailesh-vaidya (approved):
Looks good.

sachitanands (approved):

PR #253, opened 2020-09-29
EOS-13747:S3:Compress s3server log in s3 support bundle

sachitanands:
Compress s3 log directory in support bundle and then collect it .

### DIFF
--- a/scripts/s3-support-bundles/s3_bundle_generate.sh
+++ b/scripts/s3-support-bundles/s3_bundle_generate.sh
@@ -56,7 +56,6 @@ s3server_binary="/opt/seagate/cortx/s3/bin/s3server"
 s3_motr_dir="/var/motr/s3server-*"
 s3_core_dir="/var/crash"
 sys_auditlog_dir="/var/log/audit"
-s3_recovery_dir="/var/log/seagate/s3/s3recovery"
 
 # Create tmp folder with pid value to allow parallel execution
 pid_value=$$
@@ -71,6 +70,7 @@ s3_core_files="$tmp_dir/s3_core_files"
 s3_m0trace_files="$tmp_dir/s3_m0trace_files"
 first_s3_m0trace_file="$tmp_dir/first_s3_m0trace_file"
 m0trace_files_count=5
+s3_core_files_max_count=11
 
 # LDAP data
 ldap_dir="$tmp_dir/ldap"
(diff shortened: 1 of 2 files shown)


scripts/s3-support-bundles/s3_bundle_generate.sh
```diff
@@ -56,7 +56,6 @@ s3server_binary="/opt/seagate/cortx/s3/bin/s3server"
 s3_motr_dir="/var/motr/s3server-*"
 s3_core_dir="/var/crash"
 sys_auditlog_dir="/var/log/audit"
-s3_recovery_dir="/var/log/seagate/s3/s3recovery"
```
sachitanands, line 59: Recovery tool log directory is already collected as part of s3 directory (/var/log/seagate/s3) .. hence no need to collect it explicitly here.

server/request_object.cc
```diff
@@ -729,10 +729,7 @@ void RequestObject::send_response(int code, std::string body) {
   if (!is_service_req_head) {
     s3_log(S3_LOG_INFO, request_id, "Sending response as: [code:%d] [%s]\n",
            code, body.c_str());
-  } else {
```
amitwac2608, line 732: tagging @knrajnambiar76 
Rajesh added this else part few days back.

amitwac2608 (approved):
seems fine with me. Confirm with Rajesh on the else part which you have removed in this PR before merging.

server/request_object.cc
knrajnambiar76: @amitwac2608 I asked @sachitanands  to remove it, some more optimization for health check requests

bkirunge7 (approved):
looks good

sachitanands:
Pre-merge job link : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/30 (In-Progress)

scripts/s3-support-bundles/s3_bundle_generate.sh
```diff
@@ -98,19 +98,20 @@ s3server_logdir=`cat $s3server_config | grep "S3_LOG_DIR:" | cut -f2 -d: | sed -
...
 
-# Compress each s3server core file present in /var/crash directory if available
+# Compress latest s3 core with <s3_core_files_max_count> files from /var/crash directory if available
 # these compressed core file will be available in /tmp/s3_support_bundle_<pid>/s3_core_files directory
 compress_core_files(){
-  core_filename_pattern="*/core-s3server.*"
-  for file in $s3_core_dir/*
+  core_filename_pattern="core-s3server.*"
+  # get recent modified core files from directory
+  core_files_list=$(ls -tr "$s3_core_dir" | grep "$core_filename_pattern" | tail -"$s3_core_files_max_count")
```
DPG17, line 106: I believe, "ls -tr" will return list with oldest file first(on the top of list) and latest at bottom. 
When you 'tail -x' on this list, your will get last x files. This will include the latest as well. But, we want to consider old files.
I think, you may need to use 'ls -t' OR 'head x'

DPG17: Probably, you may have to use 'ls -ltr' and than use head -n x to get the 'x' oldest files

PR #260, opened 2020-09-30
EOS-13788: S3 server to send alert in case of hang in motr initializa…

knrajnambiar76:
…tion

S3 server to send IEM Alert whenever there is a hang in motr initialization

### DIFF
--- a/server/s3server.cc
+++ b/server/s3server.cc
@@ -58,6 +58,7 @@
 #define MIN_RESERVE_SIZE 32768
 
 #define WEBSTORE "/home/seagate/webstore"
+#define MOTR_INIT_MAX_ALLOWED_TIME 15
 
 /* Program options */
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


amitwac2608 (approved):

server/s3server.cc
```diff
@@ -96,6 +97,20 @@ std::set<struct s3_motr_idx_op_context *> global_motr_idx_ops_list;
...
+  s3_iem(LOG_ALERT, S3_IEM_MOTR_CONN_FAIL, S3_IEM_MOTR_CONN_FAIL_STR,
+         S3_IEM_MOTR_CONN_FAIL_JSON);
+  event_base_loopbreak(global_evbase_handle);
+  return;
+}
+
+void *base_loop_thread(void *arg) {
+  pthread_detach(pthread_self());
+  event_base_loop(global_evbase_handle, EVLOOP_NO_EXIT_ON_EMPTY);
+  return NULL;
```
DPG17, line 111: Don't we need **conditional** exit() call here to terminate S3server process when event that is set to occur after 15secs is raised.
Presently, this thread will just terminate at line #111, but it won't terminate the other thread (in the process) where motr initliazation is carried out and stuck due to initialization issue in motr. This won't bring down the entire s3server process. Will it?

DPG17: Per discussion with Rajesh, if hax(part of hare), after receiving the IEM message from s3server indicating hang in motr initialization, restarts the entire IO stack (i.e motr services + s3service), then it is ok. Otherwise, these changes won't terminate S3server service as such, when there is motr initialization issue.

DPG17 (approved):

knrajnambiar76:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/47/console

PR #263, opened 2020-10-01
EOS-13458:[LDR-R2] Generate request id (Unique id) from haproxy

AjinkyaDhumal:
Signed-off-by: Ajinkya Dhumal <ajinkya.dhumal@seagate.com>

### DIFF
--- a/auth/server/src/main/java/com/seagates3/authserver/AuthServerPostHandler.java
+++ b/auth/server/src/main/java/com/seagates3/authserver/AuthServerPostHandler.java
@@ -65,12 +65,19 @@ public class AuthServerPostHandler {
     public void run() {
         Map<String, String> requestBody = getHttpRequestBodyAsMap();
 
-        // Generate request Id per request
-        if (!(requestBody.get("Request_id") == null ||
-              (requestBody.get("Request_id")).isEmpty())) {
-          AuthServerConfig.setReqId(requestBody.get("Request_id"));
-        } else {
-          AuthServerConfig.setReqId(BinaryUtil.getAlphaNumericUUID());
+        // Check if Request_Id is present in request body
+        String reqId = null;
+        if (requestBody != null) reqId = requestBody.get("Request_id");
+        if (reqId != null && !reqId.isEmpty()) {
+          AuthServerConfig.setReqId(reqId);
+        } else {  // Else check if Request_Id is present in request header
+          if (httpRequest.headers() != null)
+            reqId = httpRequest.headers().get("Request_id");
+          if (reqId != null && !reqId.isEmpty()) {
+            AuthServerConfig.setReqId(reqId);
+          } else {  // Else generate a new ID and set
+            AuthServerConfig.setReqId(BinaryUtil.getAlphaNumericUUID());
+          }
         }
 
         if (httpRequest.getUri().startsWith("/saml")) {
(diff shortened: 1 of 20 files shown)


scripts/haproxy/haproxy_osver7.cfg
```diff
@@ -122,6 +122,11 @@ frontend main
     acl s3authbackendacl dst_port 9080
     use_backend s3-auth if s3authbackendacl
 
+    # Add a Request ID to request header and haproxy log
+    unique-id-format        %{+X}o\ %ci:%cp_%fi:%fp_%Ts_%rt:%pid
```
bkirunge7, line 126: we can drop frontend ip and port, as it will always be same on a node

server/request_object.cc
```diff
@@ -96,6 +97,36 @@ void RequestObject::listen_for_incoming_data(std::function<void()> callback,
   resume();  // resume reading if it was paused
 }
 
+static void generate_uuid(const std::string& raw_str, char uuid[16]) {
```
bkirunge7, line 100: can we use existing uuid.get_string_uuid() and trim if required

scripts/haproxy/haproxy_osver8.cfg
```diff
@@ -135,6 +135,11 @@ frontend main
     acl s3authbackendacl dst_port 9080
     use_backend s3-auth if s3authbackendacl
 
+    # Add a Request ID to request header and haproxy log
+    unique-id-format        %{+X}o\ %ci:%cp_%fi:%fp_%Ts_%rt:%pid
+    unique-id-header        Request_id
```
bkirunge7, line 140: client may also send such headers, may be we can change this to some uncommon value

scripts/haproxy/haproxy_osver7.cfg
evgeniy-brazhnikov: Will do

server/request_object.cc
evgeniy-brazhnikov: No. Another format

scripts/haproxy/haproxy_osver8.cfg
evgeniy-brazhnikov: Will be changed

knrajnambiar76 (approved):

auth/server/src/main/java/com/seagates3/authserver/AuthServerPostHandler.java
```diff
@@ -65,12 +65,19 @@ public AuthServerPostHandler(ChannelHandlerContext ctx,
...
-        } else {
-          AuthServerConfig.setReqId(BinaryUtil.getAlphaNumericUUID());
+        // Check if Request_Id is present in request body
+        String reqId = null;
+        if (requestBody != null) reqId = requestBody.get("Request_id");
+        if (reqId != null && !reqId.isEmpty()) {
+          AuthServerConfig.setReqId(reqId);
+        } else {  // Else check if Request_Id is present in request header
+          if (httpRequest.headers() != null)
+            reqId = httpRequest.headers().get("Request_id");
```
knrajnambiar76, line 75: This may now need change as 
unique-id-header        X-Unique-ID

bkirunge7 (approved):

PR #278, opened 2020-10-05
EOS-13916 Failover may cause some IOs to fail since authserver takes …

abhilekh268:
…some time to be listening to port, during this time the failed over IO to secondary node may fail


knrajnambiar76 (commented):
Abhilekh, whenever s3server fails to connect to auth server and if we have crossed the retry interval we should should return to client ServiceUnavailable

amitwac2608 (commented):
looks good.

knrajnambiar76 (approved):

nileshgovande (approved):

PR #295, opened 2020-10-09
EOS-14137: Removed fatal signal handler registration.

DPG17:

### DIFF
--- a/server/s3server.cc
+++ b/server/s3server.cc
@@ -751,8 +751,9 @@ int main(int argc, char **argv) {
   S3Daemonize s3daemon;
   set_fatal_handler_exit();
   s3daemon.daemonize();
+#if 0
   s3daemon.register_signals();
-
+#endif
   // dump the config
   g_option_instance->dump_options();
 


DPG17:
Per discussion with @knrajnambiar76, removed the registration of fatal handler with OS.
This handler is handling fatal events (SIGABRT, SIGKILL, SIGFPE, SIGSEGV)

There is no need to handle: SIGKILL (as OS will never pass this signal to the process). Rest all fata signals, with their default handler would not require any further custom handler.

nileshgovande:
@knrajnambiar76  and @DPG17 Why did we have it registered originally then?

amitwac2608 (approved):

server/s3server.cc
```diff
@@ -751,7 +751,6 @@ int main(int argc, char **argv) {
   S3Daemonize s3daemon;
   set_fatal_handler_exit();
   s3daemon.daemonize();
-  s3daemon.register_signals();
```
knrajnambiar76, line 754: you can keep this commented of in #if 0...  since other code is still in place

knrajnambiar76 (approved):

DPG17:
PR jenkins build: http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/112 is triggered.

server/s3server.cc
DPG17: ok.

DPG17:
Jenkins PR build: http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/114/ is triggered.

nileshgovande:
jenkins passed

PR #297, opened 2020-10-09
EOS-14064- preventive fix for avoiding NPE

shalakadharap:


abhilekh268 (approved):

nileshgovande (approved):

shalakadharap:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/121/
Build SUccess

auth/server/src/main/java/com/seagates3/acl/ACLValidation.java
```diff
@@ -224,7 +224,7 @@ static boolean checkIdExists(String canonicalID, String displayname) {
       return false;
     }
 
-    if (ldapResults.hasMore()) {
+    if (ldapResults != null && ldapResults.hasMore()) {
```
amitwac2608, line 227: What action to take if ldapResults is NULL ?

Should we instead of doing if check, add  **catch(NullPointerException e)** ?

amitwac2608 (changes requested):
Shalaka - I suggest to use  catch(NullPointerException e)  instead of doing if check.

and put the action needed (log or return) in the catch block.

auth/server/src/main/java/com/seagates3/acl/ACLValidation.java
shalakadharap: No. Because of this check it will return false if ldapresult is NULL and we will get expected behavior
Same is the case with all files.
Also, this change is already in R2 we just patching back in main

amitwac2608: good to hear that we are taking right action if the value is NULL.

But I still believe its better we use catch block instead of checking for NULL in if block.

amitwac2608: do you have other concern on using catch ?

amitwac2608: Also Shalaka, where exactly we are returning false if ldapResults is null ??

I see we go to line 242, if ldapResults is set to null !!! Am I missing something ?

shalakadharap: Catching NPE is not required here. and its not good practice

amitwac2608 (approved):

PR #300, opened 2020-10-10
EOS:14187: [LDR R1] handle error scenario in callback: save_object_metadata_failed

amitwac2608:
Put object was returning 200 even if motr set state to "failed", becasue
we were only checking "failed_to_launch" state in the callback. Added else
condition to return InternalError for all other states in the callback for failure.


knrajnambiar76 (approved):

amitwac2608:
Jenkins job running: http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/125/

Job 124 already passed (without UT changes.)

nileshgovande (approved):

PR #315, opened 2020-10-16
[LDR-R1]Eos14207 SW Update with sample file approach

shalakadharap:


bkirunge7 (approved):

auth/encryptcli/src/test/java/com/seagates3/authpassencryptcli/AuthEncryptCLITest.java
```diff
@@ -57,7 +57,7 @@
     @Before
     public void setUp() throws Exception {
         //For Unit Tests resource folder is at s3server/auth/resources
-        String installDir = "..";
+      String installDir = "/opt/seagate/cortx/auth";
```
AjinkyaDhumal, line 60: I think we should continue using the code files here. Better to use the sample files. Production specific references should be avoided for unit tests. It would fail for fresh dev VMs where /opt/seagate is not present.

AjinkyaDhumal (changes requested):

auth/encryptcli/src/test/java/com/seagates3/authpassencryptcli/AuthEncryptCLITest.java
shalakadharap: Even on fresh VM,  we will be performing jenkins build atleast 1s. Build will anyways copy all files to install location so should not fail

sachitanands: on fresh vm, this will fail even for first jenkins run because auth UTs will be run first and then we are creating/copying files to /opt/seagate directory.. Please refer rebuildall.sh script
For Unit test can we use source directory cortx-s3server/auth/ instead of install directory?

nileshgovande: We can create .config <- .sample softlinks in the code ?

shalakadharap: @sachitanands @nileshgovande
It will not fail on fresh vm as well.
Kindly check rebuildall.sh file changes which i checked in already.
let me know if still think is an issue

AjinkyaDhumal (approved):
Looks good.

bkirunge7 (approved):

nileshgovande:
HW test and Jenkins complete

PR #322, opened 2020-10-19
EOS-14293: add version id initialization

evgeniy-brazhnikov:

### DIFF
--- a/server/s3_object_metadata.cc
+++ b/server/s3_object_metadata.cc
@@ -670,10 +670,6 @@ int S3ObjectMetadata::from_json(std::string content) {
   account_name = system_defined_attribute["Owner-Account"];
   account_id = system_defined_attribute["Owner-Account-id"];
   object_version_id = system_defined_attribute["x-amz-version-id"];
-
-  if (object_version_id.empty()) {
-    regenerate_version_id();
-  }
   rev_epoch_version_id_key =
       S3ObjectVersioingHelper::generate_keyid_from_versionid(object_version_id);
 
--- a/server/s3_object_metadata.cc
+++ b/server/s3_object_metadata.cc
@@ -670,6 +670,10 @@ int S3ObjectMetadata::from_json(std::string content) {
   account_name = system_defined_attribute["Owner-Account"];
   account_id = system_defined_attribute["Owner-Account-id"];
   object_version_id = system_defined_attribute["x-amz-version-id"];
+
+  if (object_version_id.empty()) {
+    regenerate_version_id();
+  }
   rev_epoch_version_id_key =
       S3ObjectVersioingHelper::generate_keyid_from_versionid(object_version_id);
 
--- a/server/s3_object_versioning_helper.cc
+++ b/server/s3_object_versioning_helper.cc
@@ -54,7 +54,7 @@ std::string S3ObjectVersioingHelper::get_versionid_from_epoch_time(
 std::string S3ObjectVersioingHelper::generate_keyid_from_versionid(
     std::string versionid) {
   // add padding if requried;
-  versionid = versionid.append((3 - versionid.size() % 3) % 3, '=');
+  versionid = versionid.append((4 - (versionid.size() & 3)) & 3, '=');
   // base64 decoding of versionid
   // decoded versionid is the epoch time in milli seconds
   std::string milliseconds_since_epoch = base64_decode(versionid);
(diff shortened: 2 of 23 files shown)


knrajnambiar76 (approved):

evgeniy-brazhnikov:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/176/

evgeniy-brazhnikov:
STs finished successfully on Dev VM too

server/s3_object_metadata.cc
```diff
@@ -670,6 +670,10 @@ int S3ObjectMetadata::from_json(std::string content) {
   account_name = system_defined_attribute["Owner-Account"];
   account_id = system_defined_attribute["Owner-Account-id"];
   object_version_id = system_defined_attribute["x-amz-version-id"];
+
+  if (object_version_id.empty()) {
```
amitwac2608, line 674: when version_id could be empty ?

We saved the object metadata with empty version ID, in previous PUT object call  ? @knrajnambiar76, @DPG17  !!

server/s3_object_versioning_helper.cc
```diff
@@ -54,7 +54,7 @@ std::string S3ObjectVersioingHelper::get_versionid_from_epoch_time(
 std::string S3ObjectVersioingHelper::generate_keyid_from_versionid(
     std::string versionid) {
   // add padding if requried;
-  versionid = versionid.append((3 - versionid.size() % 3) % 3, '=');
+  versionid = versionid.append((4 - (versionid.size() & 3)) & 3, '=');
```
amitwac2608, line 57: @evgeniy-brazhnikov can you explain a little on why this change ?

server/s3_object_metadata.cc
amitwac2608: yes, but we need to understand why it was empty ? Is there some issue with PUT request !!

server/s3_object_versioning_helper.cc
evgeniy-brazhnikov: x % 4 === x & 3 for each integer x.
So formula is (4 - (versionid.size() % 4)) %4
There was the error before.
Bitwise AND is much faster then division.

server/s3_object_metadata.cc
DPG17: object_version_id, in this from_json() function, should never be empty.

DPG17: We need to investigate why object_version_id is empty.

server/s3_object_versioning_helper.cc
amitwac2608: ok.

DPG17 (changes requested):
object_version_id, in this from_json() function, should never be empty.
@evgeniy-brazhnikov : you need to investigate why object_version_id is empty.

server/s3_object_metadata.cc
evgeniy-brazhnikov: @knrajnambiar76 @DPG17 @amitwac2608 
M.b. something is incomplete while initializing multipart upload?

amitwac2608 (approved):

evgeniy-brazhnikov:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server/177/
Tests successfully finished on Dev VM also.

evgeniy-brazhnikov:
Datta demanded to undo changes related to base64. Will do

server/s3_post_multipartobject_action.cc
```diff
@@ -524,6 +524,8 @@ void S3PostMultipartObjectAction::save_upload_metadata() {
   object_multipart_metadata->set_content_type(request->get_content_type());
   object_multipart_metadata->set_part_index_oid(
       part_metadata->get_part_index_oid());
+  object_multipart_metadata->regenerate_version_id();
```
DPG17, line 527: This action class is about 'initiate multipart'. In this request, S3 server never creates object, but it is still not live to external world. When Post complete request is sent to S3 server, the object created here will be committed or will go live to external world. It is only during post complete request, the object is actually live, and the version is created during post complete action.

So, this regenerate_version_id() is not required here, as this is not final post complete action that makes the object live.

DPG17 (approved):

server/s3_post_multipartobject_action.cc
DPG17: Ok, this regenerate_version_id() here will fix assert() issue that was observed during ST run.
Post complete action also calls regenerate_version_id() when object is actually becoming live.

PR #335, opened 2020-10-22
EOS-14400 Delete expired temporary keys

shalakadharap:


auth/server/src/main/java/com/seagates3/dao/ldap/AccessKeyImpl.java
```diff
@@ -491,6 +491,59 @@ private void saveFedAccessKey(AccessKey accessKey)
...
+    @Override public void deleteExpiredKeys(User user)
+        throws DataAccessException {
+
+      String[] attrs = {LDAPUtils.ACCESS_KEY_ID, LDAPUtils.EXPIRY};
+
+      String accessKeyBaseDN =
+          String.format("%s=accesskeys,%s", LDAPUtils.ORGANIZATIONAL_UNIT_NAME,
+                        LDAPUtils.BASE_DN);
+
+      String filter = String.format("(&(%s=%s)(%s=%s))", LDAPUtils.USER_ID,
```
bkirunge7, line 506: can we add this in filter LDAPUtils.EXPIRY

auth/server/src/main/java/com/seagates3/controller/TempAuthCredentialsController.java
```diff
@@ -84,6 +84,11 @@
           if (response == null) {
             authenticateWithLdap(account, user);
             LOGGER.debug("LDAP Authentication successfull");
+            AccessKeyDAO accessKeyDAO =
+                (AccessKeyDAO)DAODispatcher.getResourceDAO(
+                    DAOResource.ACCESS_KEY);
+            // Perform cleanup before creating new temp accesskey
```
sachitanands, line 90: @shalakadharap @bkirunge7 Can we log one liner saying "Deleting access key for < account-name >"

auth/server/src/main/java/com/seagates3/dao/ldap/AccessKeyImpl.java
shalakadharap: We have stored expiry date as string and we haven't defined any search type for expiry date attribute definition.
olcAttributeTypes: {12}( 1.3.6.1.4.99999.2.3.13 NAME ( 'exp' 'expiry' ) DESC '
 Expiry TimeStamp' SYNTAX 1.3.6.1.4.1.1466.115.121.1.24 )
Because of this, as of now we cannot put exp in filter (i.e. we cannot use comparator operator on exp)
In order to use expiry in ldapsearch  query filter, 
1. we need to modify attribute definition to add 'less than equal' or 'greater than equal' type
2.we need to check if it works well with strings i.e. if it perform correct date comparison or no

auth/server/src/main/java/com/seagates3/controller/TempAuthCredentialsController.java
shalakadharap: I will log @sachitanands 
Confirmed with @bkirunge7

shalakadharap:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/32/
Jenkins success

sachitanands (approved):

bkirunge7 (approved):

PR #382, opened 2020-11-07
Added fsync call after last write_object to sync object write.

amitwac2608:

### DIFF
--- a/server/s3_motr_writer.cc
+++ b/server/s3_motr_writer.cc
@@ -372,6 +372,48 @@ void S3MotrWiter::create_object_failed() {
   s3_log(S3_LOG_DEBUG, "", "Exiting\n");
 }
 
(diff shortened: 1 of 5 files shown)


server/s3_motr_writer.cc
```diff
@@ -372,6 +372,48 @@ void S3MotrWiter::create_object_failed() {
...
+  s3_log(S3_LOG_INFO, request_id, "Entering\n");
+  handler_on_success = std::move(on_success);
+  handler_on_failed = std::move(on_failed);
+
+  struct m0_op *sync_op = NULL;
+  int rc = m0_sync_op_init(&sync_op);
+  if (rc != 0) {
+    m0_op_fini(sync_op);
+    m0_op_free(sync_op);
+  }
```
madhavemuri, line 386: If sync_op_init() fails it will run sync_entity_add(), can you change it.
Do the same after sync_entity_add().

PR #386, opened 2020-11-09
[DI POC] Added M0_OOF_SYNC flag for the last write call to motr for an object …

amitwac2608:
…put request.

### DIFF
--- a/server/s3_motr_writer.cc
+++ b/server/s3_motr_writer.cc
@@ -477,11 +477,9 @@ void S3MotrWiter::write_content(bool is_last_write) {
 
   /* Create the write request */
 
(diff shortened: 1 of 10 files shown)


server/s3_motr_writer.cc
```diff
@@ -388,15 +388,15 @@ void S3MotrWiter::write_content(
   assert(oid_list.size() == 1);
 
   if (is_object_opened) {
-    write_content();
+    write_content(is_last_write);
   } else {
     open_objects();
```
amitwac2608, line 393: I need to check if we need to pass **is_last_write** flag to open_object() as well (from open_object_successful(), we call write_content())

nileshgovande:
Closing PR as not needed for R1. The branch is still there

PR #391, opened 2020-11-10
EOS 14891 : Removed aws dependency for m0kv_metadata_parsing_tool.

SaumitraK531:
Signed-off-by: Saumitra Kulkarni <saumitra.kulkarni@seagate.com>

### DIFF
--- a/m0kv_metadata_parsing_tool/m0kv_metadata_parsing_tool.sh
+++ b/m0kv_metadata_parsing_tool/m0kv_metadata_parsing_tool.sh
@@ -26,24 +26,12 @@ reset=$(tput sgr0)
 readonly probable_delete_index='0x7800000000000000:0x100003'
 readonly bucket_list_index='0x7800000000000000:0x100002'
 
-
 rm base64_encoder_decoder -f > /dev/null 2>&1
 rm /var/log/m0kv_metadata.log > /dev/null 2>&1
 make > /dev/null 2>&1
 
 ##### Finding path of m0kv in cortx-s3server #####
-cd ..
-for line in $(find -name m0kv) 
-do
-    temp=$(echo "$line" | cut -c3-)
-    val=$(more "$temp" | grep -a ' temporary wrapper script for .libs/m0kv' ) 
-    if [[ ! -z "$val" ]]
-    then
-        m0kv_PATH=$(echo "$temp")
-    fi
-done
-cd m0kv_metadata_parsing_tool
-
+m0kv_PATH=$(find / ! -path /proc -name m0kv -type f | head -n1)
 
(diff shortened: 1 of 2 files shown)


m0kv_metadata_parsing_tool/m0kv_metadata_parsing_tool.sh
```diff
@@ -26,24 +26,12 @@ reset=$(tput sgr0)
...
-    temp=$(echo "$line" | cut -c3-)
-    val=$(more "$temp" | grep -a ' temporary wrapper script for .libs/m0kv' ) 
-    if [[ ! -z "$val" ]]
-    then
-        m0kv_PATH=$(echo "$temp")
-    fi
-done
-cd m0kv_metadata_parsing_tool
-
+m0kv_PATH=$(find / ! -path /proc -name m0kv -type f | head -n1)
```
amitwac2608, line 34: Can't we simply hard code the path, found on HW ?

m0kv_metadata_parsing_tool/m0kv_metadata_parsing_tool.sh
```diff
@@ -77,80 +65,91 @@ get_probable_delete_index() {
         declare -a PART_LIST_INDEX_OID
         rm /tmp/m0kvprobable.log > /dev/null 2>&1
         cd ..
-        ./"$m0kv_PATH" -l "$1" -h "$2" -p "$3" -f "$4" index next "$probable_delete_index" '0' 1000 -s >> /tmp/m0kvprobable.log
+        "$m0kv_PATH" -l "$1" -h "$2" -p "$3" -f "$4" index next "$probable_delete_index" '0' 1000 -s >> /tmp/m0kvprobable.log
```
amitwac2608, line 68: can we take input from user as to where exactly do we want to save the log ?

Or this log file will be deleted later by the script ?

amitwac2608 (commented):
coouple of comments. Rest looks good.

m0kv_metadata_parsing_tool/m0kv_metadata_parsing_tool.sh
SaumitraK531: No, because tool is for both dev and production environment. So if hardcode path it can fail on dev environment.

m0kv_metadata_parsing_tool/m0kv_metadata_parsing_tool.sh
SaumitraK531: Log file is stored in /var/log/

amitwac2608 (approved):

nileshgovande (approved):

nileshgovande:
Standalone tool - review complete - Not in jenkins path

PR #393, opened 2020-11-11
[LDR-R1] S3: EOS-15028: Handle invalid bucket name urlencoding

sachitanands:
invalid bucketname with special characters needs to be handled before while sending Authenication request to auth client 
if not then Auth server might fail with SignatureDoesNotMatch error

### DIFF
--- a/server/s3_auth_client.cc
+++ b/server/s3_auth_client.cc
@@ -675,11 +675,13 @@ bool S3AuthClient::setup_auth_request_body() {
   const char *full_path = request->c_get_full_encoded_path();
   if (full_path != NULL) {
     std::string uri_full_path = full_path;
-    // in encoded uri path space is encoding as '+'
-    // but cannonical request should have '%20' for verification.
-    // decode plus into space, this special handling
-    // is not required for other charcters.
-    S3CommonUtilities::find_and_replaceall(uri_full_path, "+", "%20");
+    // Handle encoding of common special characters in uri path space,
+    // if not then cannonical request generation might break with
+    // SignatureDoesNotMatch error.
+    // List of special characters handled are,
+    // '!', '"', '$', '&', ''', '(', ')', '*', '+', ',', ':', ';', '', '<',
+    // '=', '>', '*'
+    uri_full_path = S3CommonUtilities::replace_special_chars(uri_full_path);
     add_key_val_to_body("ClientAbsoluteUri", uri_full_path);
   } else {
     add_key_val_to_body("ClientAbsoluteUri", "");
--- a/server/s3_common_utilities.cc
+++ b/server/s3_common_utilities.cc
@@ -30,6 +30,34 @@
 
 namespace S3CommonUtilities {
 
+// hashtable to identify special character for replacement into encoded format
+// in canonical request
+// supported special characters are '!', '"', '$', '&', ''', '(', ')', '*',
+// '+', ',', ':', ';', '', '<', '=', '>', '@'
+// e.g '!' -> canonical_special_char[33] -> url_encoded_special_chars[33] ->
+// '%21'
+const char canonical_special_char[128] = {
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1,
+    1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 1, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
+const char *url_encoded_special_chars[128] = {
+    "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",
+    "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",   "0",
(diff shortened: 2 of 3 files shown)


server/s3_auth_client.cc
```diff
@@ -675,11 +675,33 @@ bool S3AuthClient::setup_auth_request_body() {
...
   if (full_path != NULL) {
     std::string uri_full_path = full_path;
-    // in encoded uri path space is encoding as '+'
-    // but cannonical request should have '%20' for verification.
-    // decode plus into space, this special handling
-    // is not required for other charcters.
+    // Handle encoding of special characters in uri path space,
+    // if not then cannonical request generation might break with
+    // SignatureDoesNotMatch error.
     S3CommonUtilities::find_and_replaceall(uri_full_path, "+", "%20");
```
bkirunge7, line 681: lets add a function that will replace this in one scan of uri_full_path string, this method will impact performance, as this is called for every request and add to time to first byte.

bkirunge7: create hashtable  special_char['+'] = "%20", special_char['!'] = "%21" ...
uri_len = uri_full_path.lenght();
string uri_for_canonical;
for(int i=0;i<uri_len;i++){
    if(special_char[uri_full_path[i]]){
        uri_for_canonical += special_char[uri_full_path[i]];
    } else {
       uri_for_canonical += uri_full_path[i];
    }
}

nileshgovande: @bkirunge7 - here 1 character gets replaced with 3 - so it is string replacement. We are still doing one more optimization. Pls check the newer version once

knrajnambiar76 (approved):

server/s3_auth_client.cc
bkirunge7: It's one scan, doesnt' depend on replaced char size, we ar appending new chars at end

server/s3_auth_client.cc
```diff
@@ -675,11 +675,19 @@ bool S3AuthClient::setup_auth_request_body() {
...
-    // is not required for other charcters.
-    S3CommonUtilities::find_and_replaceall(uri_full_path, "+", "%20");
+    // Handle encoding of common special characters in uri path space,
+    // if not then cannonical request generation might break with
+    // SignatureDoesNotMatch error.
+    const char *arr_syms[13] = {"+", "!", "*", "$", "'", "?", "=",
+                                "@", ":", ";", "#", "&", ","};
+    const char *arr_rep[13] = {"%20", "%21", "%2A", "%24", "%27", "%3F", "%3D",
+                               "%40", "%3A", "%3B", "%23", "%26", "%2C"};
+    for (int i = 0; i < 13; i++) {
```
bkirunge7, line 685: lets revise this code, we can avoid calling find replace multiple times

abhilekh268: @bkirunge7  this check can increase a request time bit...as every request will go through this.

server/s3_auth_client.cc
sachitanands: done .. pushed latest changes

server/s3_auth_client.cc
sachitanands: done..with latest approach we will be doing only 1 scan and then are replacing encoded chars in uri

sachitanands:
Tested basic scenarios and all splunk test went fine..

image

Pre-merge job : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/107/

sachitanands:
image

server/s3_common_utilities.cc
```diff
@@ -30,6 +30,29 @@
...
+// Special char hash table
+const char special_char[128] = {
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 1, 1, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
+const char *url_encoded_special_chars_test[128] = {
```
abhilekh268, line 42: put a comment

abhilekh268 (approved):

server/s3_common_utilities.cc
```diff
@@ -30,6 +30,29 @@
 
 namespace S3CommonUtilities {
 
+// Special char hash table
```
bkirunge7, line 33: add a comment and list all special chars handled here
// '!' -> special_char[30]
its better to make it specific to canonical id name it canonical_special_char as we are not required to handle all spcial chars like '/'

server/s3_common_utilities.cc
bkirunge7, line 42: url_encoded_special_chars_test remove test suffix, 
this table can contain mapping for all special chars encoded strings, its not required that it should match to first table

server/s3_auth_client.cc
```diff
@@ -675,11 +675,10 @@ bool S3AuthClient::setup_auth_request_body() {
   const char *full_path = request->c_get_full_encoded_path();
   if (full_path != NULL) {
     std::string uri_full_path = full_path;
-    // in encoded uri path space is encoding as '+'
-    // but cannonical request should have '%20' for verification.
-    // decode plus into space, this special handling
-    // is not required for other charcters.
-    S3CommonUtilities::find_and_replaceall(uri_full_path, "+", "%20");
+    // Handle encoding of common special characters in uri path space,
```
bkirunge7, line 678: add comment on list of special chars encoded here

server/s3_common_utilities.cc
sachitanands: done

server/s3_common_utilities.cc
sachitanands: done

sachitanands:
per-merge job : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/109/

bkirunge7 (approved):

PR #402, opened 2020-11-17
EOS-8749: Revert global indexes on create bucket error

dsurnin:

### DIFF
--- a/server/s3_bucket_metadata_v1.cc
+++ b/server/s3_bucket_metadata_v1.cc
@@ -53,6 +53,8 @@ S3BucketMetadataV1::S3BucketMetadataV1(
   bucket_owner_account_id = "";
   // name of the index which holds all objects key values within a bucket
   salted_object_list_index_name = get_object_list_index_name();
+
+  should_cleanup_global_idx = false;
 }
 
 struct m0_uint128 S3BucketMetadataV1::get_bucket_metadata_list_index_oid() {
@@ -221,7 +223,7 @@ void S3BucketMetadataV1::update(std::function<void(void)> on_success,
   this->handler_on_success = on_success;
   this->handler_on_failed = on_failed;
 
-  save_bucket_info();
+  save_bucket_info(false);
 
   s3_log(S3_LOG_DEBUG, "", "Exiting\n");
 }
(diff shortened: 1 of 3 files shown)


dsurnin:
http://eos-jenkins.colo.seagate.com/job/S3server/job/s3-github-test/89/

johnbent:
@dsurnin , are you getting the reviews you need on this?

server/s3_bucket_metadata_v1.cc
```diff
@@ -482,12 +469,17 @@ void S3BucketMetadataV1::save_replica() {
...
+  auto kv_state = motr_kv_writer->get_state();
+  if (!should_cleanup_global_idx) {
+    if (kv_state == S3MotrKVSWriterOpState::failed_to_launch) {
+      state = S3BucketMetadataState::failed_to_launch;
+    } else {
+      state = S3BucketMetadataState::failed;
+    }
+    this->handler_on_failed();
   } else {
-    state = S3BucketMetadataState::failed;
```
bkirunge7, line 488: will this state value will be used at other places

evgeniy-brazhnikov (approved):
Good for me

server/s3_bucket_metadata_v1.cc
dsurnin: yes, this is an operation status

bkirunge7 (commented):
@dsurnin see if you can add a System test for this use KV tool check if index is exists

dsurnin:
> 
> 
> @dsurnin see if you can add a System test for this use KV tool check if index is exists

I already modified existing tests and added a couple of new

bkirunge7:
> > @dsurnin see if you can add a System test for this use KV tool check if index is exists
> 
> I already modified existing tests and added a couple of new

I see unit tests added, no system tests, like create a bucket using aws cli with fault injection enabled, and see if it does cleanup in index table.  If its feasible then fine.

dsurnin:
> 
> 
> > > @dsurnin see if you can add a System test for this use KV tool check if index is exists
> > 
> > 
> > I already modified existing tests and added a couple of new
> 
> I see unit tests added, no system tests, like create a bucket using aws cli with fault injection enabled, and see if it does cleanup in index table. If its feasible then fine.

I didn't bring any new functionality, just reused old one, already proved and tested. In unit tests I checked that it is properly called. I think it should be enough.
Approach you suggest will require several server restarts - we do not use FI bucket creation now in tests, so stop prev, reconfigure startup, run single test for this particular scenario, stop and reconfigure as it was previous,start and continue testing. It will increase build time and test logic since we do not assume any restarting now.

bkirunge7 (approved):

PR #408, opened 2020-11-20
Added S3 dependency info

Saumya-Sunder:
Added S3 dependency info

Signed-off-by: Saumya Sunder <saumya.sunder@seagate.com>


nileshgovande:
not needed

PR #409, opened 2020-11-23
EOS:15088 - Load source metadata for CopyObject API

amitwac2608:
**Test cases executed:**

1. Non-Existing destination bucket (Error thrown: NoSuchBucket)
2. Non-Existing source bucket (Error thrown: NoSuchBucket)
3. Non-Existing source object (Error thrown: NoSuchKey)
4. Source Object Size greater than 5GB (Error thrown: InvalidArgument)
5. Wrong input (No separator b/w buxket name and object name) to "x-amz-copy-source" header. (Error thrown: InvalidArgument)
6. Source and Destination same (Error thrown: InvalidRequest)

### DIFF
--- a/server/s3_bucket_metadata.cc
+++ b/server/s3_bucket_metadata.cc
@@ -29,41 +29,20 @@
 #include "s3_common_utilities.h"
 #include "s3_m0_uint128_helper.h"
 
-S3BucketMetadata::S3BucketMetadata(
-    std::shared_ptr<S3RequestObject> req, std::shared_ptr<MotrAPI> motr_api,
-    std::shared_ptr<S3MotrKVSReaderFactory> motr_s3_kvs_reader_factory,
-    std::shared_ptr<S3MotrKVSWriterFactory> motr_s3_kvs_writer_factory)
-    : request(std::move(req)), json_parsing_error(false) {
-  request_id = request->get_request_id();
-  s3_log(S3_LOG_DEBUG, request_id, "Constructor");
-
+void S3BucketMetadata::initialize(std::string str_bucket_name) {
   account_name = request->get_account_name();
   account_id = request->get_account_id();
   user_name = request->get_user_name();
   owner_canonical_id = request->get_canonical_id();
   user_id = request->get_user_id();
-  bucket_name = request->get_bucket_name();
-
+  if (str_bucket_name.empty()) {
+    bucket_name = request->get_bucket_name();
+  } else {
+    bucket_name = str_bucket_name;
+  }
   state = S3BucketMetadataState::empty;
   current_op = S3BucketMetadataCurrentOp::none;
 
-  if (motr_api) {
-    s3_motr_api = std::move(motr_api);
-  } else {
-    s3_motr_api = std::make_shared<ConcreteMotrAPI>();
-  }
-
-  if (motr_s3_kvs_reader_factory) {
-    motr_kvs_reader_factory = std::move(motr_s3_kvs_reader_factory);
-  } else {
-    motr_kvs_reader_factory = std::make_shared<S3MotrKVSReaderFactory>();
-  }
-
-  if (motr_s3_kvs_writer_factory) {
-    motr_kvs_writer_factory = std::move(motr_s3_kvs_writer_factory);
-  } else {
-    motr_kvs_writer_factory = std::make_shared<S3MotrKVSWriterFactory>();
-  }
   collision_salt = "index_salt_";
   collision_attempt_count = 0;
 
@@ -87,6 +66,65 @@ S3BucketMetadata::S3BucketMetadata(
   system_defined_attribute["Owner-User-id"] = "";
   system_defined_attribute["Owner-Account"] = "";
   system_defined_attribute["Owner-Account-id"] = "";
+}
+
+S3BucketMetadata::S3BucketMetadata(
+    std::shared_ptr<S3RequestObject> req, std::shared_ptr<MotrAPI> motr_api,
+    std::shared_ptr<S3MotrKVSReaderFactory> motr_s3_kvs_reader_factory,
+    std::shared_ptr<S3MotrKVSWriterFactory> motr_s3_kvs_writer_factory)
+    : request(std::move(req)), json_parsing_error(false) {
+  request_id = request->get_request_id();
+  s3_log(S3_LOG_DEBUG, request_id, "Constructor");
+
+  initialize();
+
+  if (motr_api) {
+    s3_motr_api = std::move(motr_api);
+  } else {
+    s3_motr_api = std::make_shared<ConcreteMotrAPI>();
+  }
+
+  if (motr_s3_kvs_reader_factory) {
+    motr_kvs_reader_factory = std::move(motr_s3_kvs_reader_factory);
+  } else {
+    motr_kvs_reader_factory = std::make_shared<S3MotrKVSReaderFactory>();
+  }
+
+  if (motr_s3_kvs_writer_factory) {
+    motr_kvs_writer_factory = std::move(motr_s3_kvs_writer_factory);
+  } else {
+    motr_kvs_writer_factory = std::make_shared<S3MotrKVSWriterFactory>();
+  }
+}
+
+S3BucketMetadata::S3BucketMetadata(
+    std::shared_ptr<S3RequestObject> req, const std::string& str_bucket_name,
+    std::shared_ptr<MotrAPI> motr_api,
+    std::shared_ptr<S3MotrKVSReaderFactory> motr_s3_kvs_reader_factory,
+    std::shared_ptr<S3MotrKVSWriterFactory> motr_s3_kvs_writer_factory)
+    : request(std::move(req)), json_parsing_error(false) {
+  request_id = request->get_request_id();
+  s3_log(S3_LOG_DEBUG, request_id, "Constructor");
+
+  initialize(str_bucket_name);
+
+  if (motr_api) {
+    s3_motr_api = std::move(motr_api);
+  } else {
+    s3_motr_api = std::make_shared<ConcreteMotrAPI>();
+  }
+
+  if (motr_s3_kvs_reader_factory) {
+    motr_kvs_reader_factory = std::move(motr_s3_kvs_reader_factory);
+  } else {
+    motr_kvs_reader_factory = std::make_shared<S3MotrKVSReaderFactory>();
+  }
+
+  if (motr_s3_kvs_writer_factory) {
+    motr_kvs_writer_factory = std::move(motr_s3_kvs_writer_factory);
+  } else {
+    motr_kvs_writer_factory = std::make_shared<S3MotrKVSWriterFactory>();
+  }
 }
 
 std::string S3BucketMetadata::get_bucket_name() { return bucket_name; }
--- a/server/s3_bucket_metadata.h
+++ b/server/s3_bucket_metadata.h
@@ -110,9 +110,19 @@ class S3BucketMetadata {
   void regenerate_new_index_name(std::string base_index_name,
                                  std::string& salted_index_name);
 
+  void initialize(std::string bucket = "");
+
  public:
   S3BucketMetadata(
       std::shared_ptr<S3RequestObject> req,
+      std::shared_ptr<MotrAPI> motr_api = nullptr,
+      std::shared_ptr<S3MotrKVSReaderFactory> motr_s3_kvs_reader_factory =
+          nullptr,
+      std::shared_ptr<S3MotrKVSWriterFactory> motr_s3_kvs_writer_factory =
+          nullptr);
+
+  S3BucketMetadata(
+      std::shared_ptr<S3RequestObject> req, const std::string& bucket,
       std::shared_ptr<MotrAPI> motr_api = nullptr,
       std::shared_ptr<S3MotrKVSReaderFactory> motr_s3_kvs_reader_factory =
           nullptr,
--- a/server/s3_copy_object_action.cc
+++ b/server/s3_copy_object_action.cc
@@ -41,7 +41,7 @@ S3CopyObjectAction::S3CopyObjectAction(
   s3_log(S3_LOG_INFO, request_id,
          "S3 API: CopyObject. Destination: [%s], Source: [%s]\n",
          request->get_object_uri().c_str(),
-         request->get_copy_object_source().c_str());
+         request->get_headers_copysource().c_str());
   s3_copy_action_state = S3CopyObjectActionState::empty;
 
   old_object_oid = {0ULL, 0ULL};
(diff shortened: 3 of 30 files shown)


server/s3_copy_object_action.cc
```diff
@@ -47,6 +47,8 @@ S3CopyObjectAction::S3CopyObjectAction(
   old_object_oid = {0ULL, 0ULL};
   old_layout_id = -1;
   new_object_oid = {0ULL, 0ULL};
+  source_bucket_name = "";
```
evgeniy-brazhnikov, line 50: It's redundant. Default constructor of std::string initialize it as empty string.

server/s3_copy_object_action.cc
```diff
@@ -82,6 +84,145 @@ void S3CopyObjectAction::setup_steps() {
...
 }
 
+void S3CopyObjectAction::get_source_bucket_and_object() {
+  s3_log(S3_LOG_DEBUG, request_id, "Entering\n");
+  std::string source = request->get_copy_object_source();
+  size_t separator_pos = source.find("/");
+  if (separator_pos != std::string::npos) {
+    source_bucket_name = source.substr(0, separator_pos);
+    source_object_name = source.substr(separator_pos + 1);
+  }
```
evgeniy-brazhnikov, line 94: Should we handle incorrect source "path"?

server/s3_copy_object_action.cc
```diff
@@ -82,6 +84,145 @@ void S3CopyObjectAction::setup_steps() {
...
+  s3_log(S3_LOG_DEBUG, request_id, "Exiting\n");
+}
+
+void S3CopyObjectAction::fetch_source_object_info_success() {
+  s3_log(S3_LOG_INFO, request_id, "Entering\n");
+  s3_log(S3_LOG_DEBUG, request_id,
+         "Successfully fetched source object metadata\n");
+
+  if (MaxCopyObjectSourceSize < source_object_metadata->get_content_length()) {
```
evgeniy-brazhnikov, line 182: Do we really need this constraint?

server/s3_copy_object_action.cc
```diff
@@ -82,6 +84,145 @@ void S3CopyObjectAction::setup_steps() {
...
+    s3_log(S3_LOG_ERROR, request_id,
+           "Source object metadata load operation failed due to pre launch "
+           "failure\n");
+    set_s3_error("ServiceUnavailable");
+  } else {
+    s3_log(S3_LOG_DEBUG, request_id, "Source object metadata fetch failed\n");
+    set_s3_error("InternalError");
+  }
+  send_response_to_s3_client();
+  s3_log(S3_LOG_DEBUG, request_id, "Exeting\n");
```
evgeniy-brazhnikov, line 209: I suspect a bug here.
"this" instance can be already destroyed at this point.
So this->request_id can point to garbage.

server/s3_copy_object_action.h
```diff
@@ -38,6 +38,8 @@ enum class S3CopyObjectActionState {
   completed,                // All stages done completely
 };
 
+const u_long MaxCopyObjectSourceSize = 5368709120;  // 5GB
```
evgeniy-brazhnikov, line 41: I would recommend 
#define MaxCopyObjectSourceSize 5368709120ull
Dont insist

server/s3_factory.h
```diff
@@ -41,23 +41,23 @@ class S3BucketMetadataFactory {
  public:
   virtual ~S3BucketMetadataFactory() {}
   virtual std::shared_ptr<S3BucketMetadata> create_bucket_metadata_obj(
-      std::shared_ptr<S3RequestObject> req) {
+      std::shared_ptr<S3RequestObject> req, std::string bucket = "") {
```
evgeniy-brazhnikov, line 44: Did you update appropriate UTs?
MOCK_METHODn ->MOCK_METHOD(n+1)

server/s3_copy_object_action.cc
amitwac2608: Yes, it is handled in validate_copyobject_request(), from where this function is being called.

server/s3_copy_object_action.cc
amitwac2608: yes, CopyObject API only support source objects of size less than 5GB

server/s3_factory.h
amitwac2608: Yes

server/s3_copy_object_action.cc
amitwac2608: Fixed.

server/s3_copy_object_action.cc
amitwac2608: Fixed.

server/s3_copy_object_action.h
amitwac2608: I generally avoid using per-processor macros, they won't go in compiler's symbols table, and hence would be seen as a number, instead of a human readable name, when we will be debugging any core-dump in future.

evgeniy-brazhnikov: 1. The standard allows compilers to treat 'const' variable of build-in type, initialized by a constant expression, as 'constexpr'. Highly likely a compiler optimizes out this variable.
2. As far as i know integer numbers are treated as int type by default.
I'm afraid that a compiler can cut the number to 1GB.
Can you add 'ul' suffix to the number.

evgeniy-brazhnikov: Checked via trivial program. gcc doesn't cut the variable.

evgeniy-brazhnikov (approved):

server/s3_copy_object_action.h
amitwac2608: Sure I will change the suffix to UL.

Actually my comment was w.r.t to "#define" macro. As this is pre-proccessor macro, the macro would be replaced by its value when the code reaches compiler.

server/s3_bucket_metadata.cc
```diff
@@ -30,7 +30,8 @@
 #include "s3_m0_uint128_helper.h"
 
 S3BucketMetadata::S3BucketMetadata(
-    std::shared_ptr<S3RequestObject> req, std::shared_ptr<MotrAPI> motr_api,
+    std::shared_ptr<S3RequestObject> req, std::string bucket,
```
DPG17, line 33: I don't understand why we pass 'std::string bucket' to S3BucketMetadata's constructor.
You can always get bucket name from RequestObject class. 
As CopyObject involves source and destination buckets, and both these are part of the request, these bucket names should be part of "RequestObject" or "S3RequestObject" class. You may add method (get/set) in S3RequestObject class to encapsulate the second bucket name.

server/s3_copy_object_action.cc
```diff
@@ -82,6 +82,158 @@ void S3CopyObjectAction::setup_steps() {
   ACTION_TASK_ADD(S3CopyObjectAction::send_response_to_s3_client, this);
 }
 
+void S3CopyObjectAction::get_source_bucket_and_object() {
```
abhilekh268, line 85: hi amit...we need to talk on this function where it should be,,,because for authorization also we will need this bucket name and object name...but this function here is after authorization.

server/s3_global_bucket_index_metadata.cc
```diff
@@ -29,7 +29,8 @@ extern struct m0_uint128 global_bucket_list_index_oid;
 extern struct m0_uint128 replica_global_bucket_list_index_oid;
 
 S3GlobalBucketIndexMetadata::S3GlobalBucketIndexMetadata(
-    std::shared_ptr<S3RequestObject> req, std::shared_ptr<MotrAPI> motr_api,
+    std::shared_ptr<S3RequestObject> req, std::string bucket,
```
DPG17, line 32: Same comment here. What is the reason to add new parameter (i.e. bucket name) to constructor?

server/s3_bucket_metadata.cc
```diff
@@ -42,7 +43,11 @@ S3BucketMetadata::S3BucketMetadata(
   user_name = request->get_user_name();
   owner_canonical_id = request->get_canonical_id();
   user_id = request->get_user_id();
-  bucket_name = request->get_bucket_name();
+  if (bucket.empty()) {
+    bucket_name = request->get_bucket_name();
+  } else {
+    bucket_name = bucket;
+  }
```
DPG17, line 50: We would not need this. I'm missing here what is the need for passing 'bucket' parameter to constructor. Similar comment above. Can you justify this need?

server/s3_copy_object_action.cc
DPG17: @amitwac2608 : I think, you may have to override 'S3Action::load_metadata() in your new action class to load all metadata about destination bucket and object, and source bucket (all the data that you need for CopyObject operation).

server/s3_bucket_metadata.cc
amitwac2608: Yes Datta you are right, I could have updated S3RequestObject class to have 2 new member variables - source_bucket and source object.
And I have no strong counter point as to why I have not done this.

My thought process was that as this is Bucket metadata class, the constructor should also be able to load metadata of a given bucket, instead of just relying on the S3RequestObject class' encapsulated bucket.

May be in future, if there arises a need for some API (new or existing) to load metadata of a (or more) given bucket(s) (other than in S3RequestObject class), then we can easily load its metadata with current changes. 
I could have written one more constructor to add this functionality, but that could have lead to code duplication.

1 more point in favor of my argument is - that for authorization, these changes would reduce code changes for authserver (after talks with @abhilekh268 )

server/s3_copy_object_action.cc
amitwac2608: I have also added a new function in RequestObject class to read source bucket and object. Authserver can use that

server/s3_bucket_metadata.cc
amitwac2608: After discussion with Datta, We have decided to not add member variables to S3RequestObject, as is suggested above.

We just want opinion from @knrajnambiar76, as to change the signature of the current constructor (as is done in above changes) or add new constructor, which would take bucket name as parameter, instead of S3RequestObject.

Can you comment @knrajnambiar76 here your opinion ?

amitwac2608: 2nd option suggested by @DPG17 , to make temporary copy of S3RequestObject class in the Action class, and set the bucket name to source bucket name. This way we do not need to change signature of bucket metadata class.

server/s3_copy_object_action.cc
```diff
@@ -82,6 +82,158 @@ void S3CopyObjectAction::setup_steps() {
...
+void S3CopyObjectAction::validate_copyobject_request() {
+  s3_log(S3_LOG_INFO, request_id, "Entering\n");
+  get_source_bucket_and_object();
+
+  if (source_bucket_name.empty() || source_object_name.empty()) {
+    set_s3_error("InvalidArgument");
+    send_response_to_s3_client();
+  } else if (if_source_and_destination_same()) {
+    s3_copy_action_state = S3CopyObjectActionState::validationFailed;
+    set_s3_error("InvalidRequest");
```
knrajnambiar76, line 229: As per current s3_error_messages.json file
  "InvalidRequest": {
    "Description": "Specifying both Canned ACLs and Header Grants is not allowed",
    "httpcode": 400
  },
The Description seems to be incorrect there

server/s3_copy_object_action.cc
```diff
@@ -181,9 +328,20 @@ void S3CopyObjectAction::send_response_to_s3_client() {
       (is_error_state() && !get_s3_error_code().empty())) {
     // Metadata saved for object is always a success condition.
     assert(s3_copy_action_state != S3CopyObjectActionState::metadataSaved);
+    S3Error error(get_s3_error_code(), request->get_request_id());
+
+    if (S3CopyObjectActionState::validationFailed == s3_copy_action_state &&
```
knrajnambiar76, line 333: We cannot have this state at all right, in validation itself we will bail out right

server/s3_copy_object_action.h
```diff
@@ -38,6 +38,15 @@ enum class S3CopyObjectActionState {
   completed,                // All stages done completely
 };
 
+const u_long MaxCopyObjectSourceSize = 5368709120ULL;  // 5GB
```
knrajnambiar76, line 41: you can have this as macro defined in cc

server/s3_global_bucket_index_metadata.cc
knrajnambiar76, line 32: overload it

server/s3_object_metadata.cc
```diff
@@ -35,15 +35,24 @@
 
 extern struct m0_uint128 global_instance_id;
 
-void S3ObjectMetadata::initialize(bool ismultipart, std::string uploadid) {
```
knrajnambiar76, line 38: you can have this renamed to initialize_destination

server/s3_object_metadata.cc
```diff
@@ -94,15 +103,15 @@ void S3ObjectMetadata::initialize(bool ismultipart, std::string uploadid) {
 
 S3ObjectMetadata::S3ObjectMetadata(
```
knrajnambiar76, line 104: In te overloaded constructor you can call initialize_source_destination() by passing source bucket and object name

server/s3_bucket_metadata.cc
```diff
@@ -30,7 +30,8 @@
 #include "s3_m0_uint128_helper.h"
 
 S3BucketMetadata::S3BucketMetadata(
```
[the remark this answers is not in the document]
DPG17: @amitwac2608: I agree with @knrajnambiar76. Overloaded constructor would be better, as you suggested during our discussion.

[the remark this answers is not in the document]
amitwac2608: Will do.

server/s3_bucket_metadata.cc
amitwac2608: I added a new constructor, which accepts bucket name, along with other parameters.

server/s3_bucket_metadata.cc
```diff
@@ -42,7 +43,11 @@ S3BucketMetadata::S3BucketMetadata(
   user_name = request->get_user_name();
   owner_canonical_id = request->get_canonical_id();
   user_id = request->get_user_id();
-  bucket_name = request->get_bucket_name();
+  if (bucket.empty()) {
+    bucket_name = request->get_bucket_name();
```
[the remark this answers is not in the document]
amitwac2608: this bucket metadata class is also used for GET bucket or GET object action, so renaming the bucket_name to destination_bucket_name would sound odd with GET Action. I think we can keep the member variable as is. Do you agree ?

server/s3_bucket_metadata.cc
amitwac2608: yes removed this.

server/s3_bucket_metadata.h
```diff
@@ -112,7 +112,7 @@ class S3BucketMetadata {
 
  public:
   S3BucketMetadata(
```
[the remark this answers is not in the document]
amitwac2608: Done.

server/s3_copy_object_action.cc
```diff
@@ -82,6 +82,158 @@ void S3CopyObjectAction::setup_steps() {
   ACTION_TASK_ADD(S3CopyObjectAction::send_response_to_s3_client, this);
 }
 
+void S3CopyObjectAction::get_source_bucket_and_object() {
+  s3_log(S3_LOG_DEBUG, request_id, "Entering\n");
+  std::string source = request->get_copy_object_source();
```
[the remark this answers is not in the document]
amitwac2608: Done

server/s3_copy_object_action.cc
amitwac2608: I have handled this in the  S3CopyObjectAction class send_response_to_s3_client().

There I set the error message according to the CopyObject action error.

server/s3_copy_object_action.cc
amitwac2608: yes, in validate_copyobject_request() function, I set the state as "validationFailed", s3 error code to "InvalidRequest." and call send_response_to_s3_client().

I have added this check here to see if s3 error code is "InvalidRequest", so set the error message accordingly.

server/s3_copy_object_action.h
amitwac2608: I avoid using pre-processor macros to define constants, as that is not good coding practice these days.

For reference: https://rules.sonarsource.com/cpp/RSPEC-5028

Do you have something else in mind where macro would be better instead of const expression ?

server/s3_global_bucket_index_metadata.cc
amitwac2608: fixed this now.

amitwac2608: done

server/s3_object_metadata.cc
amitwac2608: I think to keep this as is, as this is generic class, and is consumed with GET action classes as well !!! Do you agree ?

server/s3_object_metadata.cc
amitwac2608: I haavr taken a different approach, I am initializing the bucket_name and object_name in constructor itslef, and not in initialize() finction. This way I am using same initialize() function in both the constructors. Do check the new code and let me know your view.

amitwac2608:
Jenkins Job: http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/178

amitwac2608:
> 
> 
> Jenkins Job: http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/178

Job PASSED

server/s3_bucket_metadata.cc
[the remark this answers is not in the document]
knrajnambiar76: Ok

knrajnambiar76 (approved):

PR #416, opened 2020-11-30
EOS-15332: S3 unable to URL decode 'continuation-token'.

DPG17:
During validation of List Object V2 request, S3 server is found to be not removing the URL encoded characters (e.g, %2F, %2B) from  'continuation-token'. The reason is unknown. Due to this, there was no removal of '%2F' and '%2B' characters from 'continuation-token', leading to further failure in base64 decoding of 'continuation-token', and thus, the object listing code enumerates from start key of the bucket, instead of from 'continuation-token'
 - Added code to explicitly perform URL decoding of 'continuation-token'.

### DIFF
--- a/server/s3_object_list_response.h
+++ b/server/s3_object_list_response.h
@@ -83,7 +83,7 @@ class S3ObjectListResponse {
   void set_max_uploads(std::string count);
   void set_max_parts(std::string count);
   void set_response_is_truncated(bool flag);
-  void set_next_marker_key(std::string next);
+  void set_next_marker_key(std::string next, bool url_encode = true);
   void set_next_marker_uploadid(std::string next);
   std::string& get_object_name();
   bool is_response_truncated() { return response_is_truncated; }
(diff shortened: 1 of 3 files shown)


DPG17:
Pre merge job triggered: http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/182/  (in progress)

knrajnambiar76 (approved):

server/s3_object_list_response.h
```diff
@@ -94,6 +94,7 @@ class S3ObjectListResponse {
     }
     return keys;
   }
+  std::string get_encoding_type() { return encoding_type; }
```
nileshgovande, line 97: Not needed

nileshgovande: Let's also write ST

nileshgovande: In next sprint

nileshgovande (approved):

PR #418, opened 2020-11-30
EOS-14918 : Key format for delayed delete record

pranavPawar07:
Signed-off-by: Pranav <pranav.pawar@seagate.com>


server/s3_put_multiobject_action.cc
```diff
@@ -575,6 +575,7 @@ void S3PutMultiObjectAction::save_metadata() {
   part_metadata->save(
       std::bind(&S3PutMultiObjectAction::next, this),
       std::bind(&S3PutMultiObjectAction::save_metadata_failed, this));
+
```
knrajnambiar76, line 578: remove this extra newline

server/s3_common_utilities.cc
```diff
@@ -126,6 +126,39 @@ bool stoi(const std::string &str, int &value) {
   return isvalid;
 }
 
+void size_based_bucketing_of_objects(std::string &oid_str, size_t obj_size){
+  s3_log(S3_LOG_INFO, ""  , "Entering\n");
+  size_t object_size_for_bucketing = obj_size;
+
```
knrajnambiar76, line 132: good to have a comment at begining, where you mention range and prefix
Object size.                                           Prefix
========                                          =====
 less than  1K                                          I
 less than 10KB and ...                           H

Also mention that this is done as listing of key's in motr happens in lexographic order, hence bigger object will get listed first.

knrajnambiar76 (approved):

server/s3_common_utilities.cc
pranavPawar07: Done.

server/s3_common_utilities.cc
```diff
@@ -126,6 +126,52 @@ bool stoi(const std::string &str, int &value) {
...
+  // 1KB < size <= 10KB                           'H'
+  // size <= 1KB                                  'I'
+
+  // NOTE : chars A/B/C have been left out for objects bigger than 100GB
+  // Listing of keys in motr happens in lexicographic order, hence bigger object
+  // will get listed first.
+
+  switch(object_size_for_bucketing){
+       case 0 ... 1024:
+            oid_str = 'I' + oid_str;      // less than or equal to 1KB
```
amitwac2608, line 148: better to use string::apend() function instead of assignment operator, that would be more efficient.

server/s3_common_utilities.cc
```diff
@@ -126,6 +126,52 @@ bool stoi(const std::string &str, int &value) {
...
+  // 50MB < size <= 1GB                           'F'
+  // 10KB < size <= 50MB                          'G'
+  // 1KB < size <= 10KB                           'H'
+  // size <= 1KB                                  'I'
+
+  // NOTE : chars A/B/C have been left out for objects bigger than 100GB
+  // Listing of keys in motr happens in lexicographic order, hence bigger object
+  // will get listed first.
+
+  switch(object_size_for_bucketing){
```
amitwac2608, line 146: Suggestion: My personal experience is switch statements are error prone.

Can we use if-else statements ? 

Like:
if (object_size_for_bucketing > 100GB) {
} else if (object_size_for_bucketing > 1GB && object_size_for_bucketing < 100GB) {
} else if () {

server/s3_common_utilities.h
```diff
@@ -44,6 +44,8 @@ class S3XORObfuscator : public S3Obfuscator {
   std::string virtual decode(const std::string &input);
 };
 
+//delayed delete
+void size_based_bucketing_of_objects(std::string &oid_str, size_t obj_size);
```
amitwac2608, line 48: const size_t obj_size !!

server/s3_common_utilities.cc
pranavPawar07: I think it is better to use switch case since it is more readable as well as if there are more than five cases in a switch statements, it is implemented using a lookup table/hash list whereas in case of 'if:s' where the last item takes much more time to reach as it has to evaluate every previous condition first.

server/s3_put_object_action.cc
```diff
@@ -623,6 +626,10 @@ void S3PutObjectAction::add_object_oid_to_probable_dead_oid_list() {
 
     // key = oldoid + "-" + newoid
     std::string old_oid_rec_key = old_oid_str + '-' + new_oid_str;
```
shanbhagsv1, line 628: For overwrite key format is <old_oid>-<new_oid>. If we expecting size based policy for both objects expected format should be

'I'<old_id>-'H'<new_oid>

Currently, it's resulting in 
'I'<old_oid>-<new_oid>

server/s3_put_object_action.cc
```diff
@@ -623,6 +626,10 @@ void S3PutObjectAction::add_object_oid_to_probable_dead_oid_list() {
...
+
+    //prepending a char depending on the size of the object (size based bucketing of object)
+    S3CommonUtilities::size_based_bucketing_of_objects(old_oid_rec_key, object_metadata->get_content_length());
```
shanbhagsv1, line 631: This size_based_bucketing_of_objects should be only for object oid and not for key.

server/s3_put_object_action.cc
```diff
@@ -635,7 +642,10 @@ void S3PutObjectAction::add_object_oid_to_probable_dead_oid_list() {
 
     probable_oid_list[old_oid_rec_key] = old_probable_del_rec->to_json();
   }
-
+  
+  //prepending a char depending on the size of the object (size based bucketing of object)
```
shanbhagsv1, line 646: This call should be before generating backgroundelete key line i.e before 628

server/s3_common_utilities.cc
```diff
@@ -126,6 +126,59 @@ bool stoi(const std::string &str, int &value) {
   return isvalid;
 }
 
+void size_based_bucketing_of_objects(std::string &oid_str, const size_t obj_size){
+  s3_log(S3_LOG_INFO, ""  , "Entering\n");
+  size_t object_size_for_bucketing = obj_size;
+  std::string temp = "";
```
shanbhagsv1, line 132: Use proper names maybe marker??

amitwac2608 (approved):
Please make sure you add UTs later.

nileshgovande:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/190/

PR #427, opened 2020-12-03
EOS-15434: ListObjectV2 listing objects even when the specified 'expected-bucket-owner' does not match.

DPG17:
Added code to handle 'x-amz-expected-bucket-owner' header.
When the specified 'x-amz-expected-bucket-owner' does not match the owner's account ID, access denied is returned.


server/s3_get_bucket_action_v2.cc
```diff
@@ -63,6 +63,24 @@ void S3GetBucketActionV2::validate_request() {
 
   std::shared_ptr<S3ObjectListResponseV2> obj_v2_list =
       std::dynamic_pointer_cast<S3ObjectListResponseV2>(object_list);
+  std::string expected_owner;
+  expected_owner = request->get_header_value("x-amz-expected-bucket-owner");
```
amitwac2608, line 67: Other APIs, like put object also support this header. Should we add support for this header to other APIs as well ? Will adding this header to just one API, not make inconsistency ?

server/s3_get_bucket_action_v2.cc
```diff
@@ -63,6 +63,24 @@ void S3GetBucketActionV2::validate_request() {
 
   std::shared_ptr<S3ObjectListResponseV2> obj_v2_list =
       std::dynamic_pointer_cast<S3ObjectListResponseV2>(object_list);
+  std::string expected_owner;
```
amitwac2608, line 66: why not initialize expected_owner here only ? we can combine line 66 and 67 !

amitwac2608 (commented):
can we add system test for this ?

server/s3_get_bucket_action_v2.cc
DPG17: ok

server/s3_get_bucket_action_v2.cc
DPG17: This may be a general Amz specific header. However, I doubt if it mentioned in PUT object api documentation (see this: https://docs.aws.amazon.com/cli/latest/reference/s3api/put-object.html)
Also, this bug is specific to object listing v2, where this option is exposed to end user. I'm not sure if it is exposed in other S3 API. Even if they are exposed, they needs to be accesses to have similar changes. Certainly, not in this change request.

DPG17: If 'expected-bucket-owner' is also required for other APIs, can you raise a bug?

knrajnambiar76 (approved):

PR #434, opened 2020-12-08
EOS-15381: copy object, major functionality

evgeniy-brazhnikov:


server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+          std::bind(&S3CopyObjectAction::read_data_block_failed, this))) {
+
+    s3_log(S3_LOG_DEBUG, request_id, "Data read is launched");
+    read_in_progress = true;
+    return;
+  }
+  if (motr_reader->get_state() == S3MotrReaderOpState::failed_to_launch) {
     set_s3_error("ServiceUnavailable");
+    s3_log(S3_LOG_ERROR, request_id,
+           "read_object_data() failed due to motr_entity_open failure\n");
```
amitwac2608, line 269: do not mention function name, just log that source object read failed

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
 }
 
-// Destination object
-void S3CopyObjectAction::fetch_object_info_success() {
-  s3_log(S3_LOG_INFO, request_id, "Entering\n");
-  next();
-  s3_log(S3_LOG_DEBUG, "", "Exiting\n");
+void S3CopyObjectAction::read_data_block_failed() {
+  assert(read_in_progress);
+  s3_put_action_state = S3PutObjectActionState::writeFailed;
```
amitwac2608, line 296: this is read function.. state should be readFailed ?

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+
+  s3_log(S3_LOG_DEBUG, request_id, "Failed to read object data from motr");
```
amitwac2608, line 298: S3_LOG_ERROR ?

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+  set_s3_error("InternalError");
+
+  if (write_in_progress) {
+    copy_failed = true;
```
amitwac2608, line 302: why not set copy_failed to true, irrespective of whether write_in_progress is true or false ?

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
 
-// Destination object
-void S3CopyObjectAction::fetch_object_info_failed() {
-  s3_log(S3_LOG_INFO, request_id, "Entering\n");
-  next();
-  s3_log(S3_LOG_DEBUG, "", "Exiting\n");
+void S3CopyObjectAction::read_data_block_success() {
+  assert(read_in_progress);
+
+  s3_log(S3_LOG_INFO, request_id, "Read object data succeeded");
```
amitwac2608, line 281: this log information is misleading... we may have not read whole object. Can we log x bytes read from total_bytes ?

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+    next();
+  }
+  s3_log(S3_LOG_DEBUG, NULL, "Exiting\n");
+}
+
+void S3CopyObjectAction::write_data_block_failed() {
+  assert(write_in_progress);
+  s3_put_action_state = S3PutObjectActionState::writeFailed;
+
+  s3_log(S3_LOG_INFO, request_id, "Failed to write object data to motr");
```
amitwac2608, line 375: S3_LOG_ERROR

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+  new_object_metadata->set_content_length(std::to_string(total_data_to_stream));
+  new_object_metadata->set_content_type(
+      source_object_metadata->get_content_type());
+  new_object_metadata->set_md5(motr_writer->get_content_md5());
+  // new_object_metadata->set_tags(new_object_tags_map);
+
+  for (auto it : request->get_in_headers_copy()) {
+    if (it.first.find("x-amz-meta-") != std::string::npos) {
+      s3_log(S3_LOG_DEBUG, request_id,
+             "Writing user metadata on object: [%s] -> [%s]\n",
```
amitwac2608, line 402: "writing source object metadata on destination object.." ?

server/s3_object_metadata.cc
```diff
@@ -35,126 +37,124 @@
...
 extern struct m0_uint128 global_instance_id;
 
-void S3ObjectMetadata::initialize(bool ismultipart, std::string uploadid) {
-  json_parsing_error = false;
+S3ObjectMetadata::S3ObjectMetadata(const S3ObjectMetadata& src)
+    : S3ObjectMetadataCopyable(src), state(S3ObjectMetadataState::present) {
+
+  s3_log(S3_LOG_DEBUG, request_id, "Copy constructor");
+
+  bucket_name = request->get_bucket_name();
```
amitwac2608, line 45: if its copy constructor, why not read bucket name and object name from the passed object itself ?

server/s3_object_metadata.cc
```diff
@@ -35,126 +37,124 @@
...
   }
-
-  object_list_index_oid = {0ULL, 0ULL};
-  objects_version_list_index_oid = {0ULL, 0ULL};
 }
 
 S3ObjectMetadata::S3ObjectMetadata(
-    std::shared_ptr<S3RequestObject> req, bool ismultipart,
-    std::string uploadid,
+    std::shared_ptr<S3RequestObject> req, std::string bucketname,
```
amitwac2608, line 107: pass the bucket name and object name as const references.

server/s3_put_object_action_base.h
```diff
@@ -0,0 +1,124 @@
...
+#include <cstddef>
+#include <memory>
+#include <string>
+
+#include <gtest/gtest_prod.h>
+
+#include "lib/types.h"  // struct m0_uint128
+#include "s3_object_action_base.h"
+
+enum class S3PutObjectActionState {
```
amitwac2608, line 31: should we rename this enum class as S3ObjectActionState, so that it would be generic and can be sued in any Action class ?

amitwac2608 (commented):
I reviewed only

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Metadata attributes are being overwritten here according HTTP headers of the request.

server/s3_object_metadata.cc
evgeniy-brazhnikov: Partially agree. Quite strange copy constructor. I thought about separate function for updating partial copy. If you insist i will do this.

server/s3_object_metadata.cc
evgeniy-brazhnikov: I thought about it. Passing of objects is less expensive in our case.

server/s3_put_object_action_base.h
evgeniy-brazhnikov: Not sure. This base class can be applied to 4 actions: put object, put multipart, put chuck and copy objects.
All are PUT. So i decided to name the enum as is.

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Such member is absent in enum. So i had to use existing.

server/s3_object_metadata.cc
amitwac2608: No I do not insist, just a suggestion to make sure we are not doing this by mistake.

If this is conscious decision, then its fine. Make sure we write good UTs to catch any error in logic.

server/s3_object_metadata.cc
amitwac2608: I got your point, but still its good practice to pass strings as const references, unless we want to update them in the function. 

We are open source, so following standard coding practices will give our code good reputation among open-source community.

server/s3_copy_object_action.cc
amitwac2608: Ok ok.. I got your point... but we need to read a header: **x-amz-metadata-directive** 

This header can have two values:
COPY (default)
REPLACE.
As of now in milestone 1, we are only supporting COPY value.

amitwac2608: With COPY, all source metadata is copied to destination object.

with REPLACE value, user can give metadata with copy-object request, and the server will replace metadata on destination with passed metadata values

evgeniy-brazhnikov: Clear. Thanks! Will update.

server/s3_copy_object_action.cc
```diff
@@ -122,7 +110,7 @@ void S3CopyObjectAction::fetch_source_bucket_info_success() {
 void S3CopyObjectAction::fetch_source_bucket_info_failed() {
   s3_log(S3_LOG_INFO, request_id, "Entering\n");
 
-  s3_copy_action_state = S3CopyObjectActionState::validationFailed;
+  s3_put_action_state = S3PutObjectActionState::validationFailed;
```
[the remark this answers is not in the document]
evgeniy-brazhnikov: This comment is correlated with the last one

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+      S3Option::get_instance()->get_motr_units_per_request();
+  const size_t rest_blocks =
+      (bytes_left_to_read + motr_unit_size - 1) / motr_unit_size;
+
+  if (motr_reader->read_object_data(
+          std::min(rest_blocks, max_blocks_in_one_read_op),
+          std::bind(&S3CopyObjectAction::read_data_block_success, this),
+          std::bind(&S3CopyObjectAction::read_data_block_failed, this))) {
+
+    s3_log(S3_LOG_DEBUG, request_id, "Data read is launched");
```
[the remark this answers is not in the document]
evgeniy-brazhnikov: Done

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Done

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Done

server/s3_object_metadata.cc
evgeniy-brazhnikov: Here the strings are passed by value, not by non-const reference.
Amit, trust me here, I exactly know what I've done.

server/s3_copy_object_action.cc
evgeniy-brazhnikov: There is no readFailed member in the enum. Also members correspond to workflow stages, not to lower level operations. It seems that "writeFailed" is quite correct.

evgeniy-brazhnikov:
jenkins-build.sh finished successfully on Dev VM

evgeniy-brazhnikov:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/207/

evgeniy-brazhnikov:
Another UTs will be added with next PR

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+  S3BufferSequence buffer_sequence;
+
+  char* p_data;
+  size_t block_size = motr_reader->get_first_block(&p_data);
+
+  unsigned block_in_chunk = 0;
+  size_t bytes_in_chunk = 0;
+
+  while (block_size) {
+    assert(p_data != NULL);
```
bkirunge7, line 324: we may need check shutdown signal here,

server/s3_copy_object_action.cc
```diff
@@ -234,74 +223,214 @@ void S3CopyObjectAction::validate_copyobject_request() {
...
+
+  if (write_in_progress) {
+    copy_failed = true;
+  } else {
+    send_response_to_s3_client();
+  }
+  s3_log(S3_LOG_DEBUG, NULL, "Exiting\n");
+}
+
+void S3CopyObjectAction::write_data_block() {
```
bkirunge7, line 309: Can we move this logic to a separate class that can be reused, and provide interface copy(oid1, oid2)

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Already added in current changes

knrajnambiar76 (approved):

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Basav, let us do it after testing basic functionality.
We must be sure that it works correctly before inserting any optimizations.
Optimization can complicate error seeking too much.

server/s3_object_metadata.cc
```diff
@@ -35,126 +37,123 @@
...
-      "";
-  system_defined_attribute["x-amz-server-side-encryption-customer-key"] = "";
-  system_defined_attribute["x-amz-server-side-encryption-customer-key-MD5"] =
-      "";
+  str_set_default(system_defined_attribute["x-amz-server-side-encryption"],
+                  "None");
+  (void)system_defined_attribute["x-amz-version-id"];
+
+  // Generate if versioning enabled
+  str_set_default(system_defined_attribute["x-amz-storage-class"], "STANDARD");
```
knrajnambiar76, line 85: why these functions ?
instead of explicit initialization

evgeniy-brazhnikov: This function is called for both empty instance and copied instance.
We must avoid overwriting copied values.

evgeniy-brazhnikov: std::map<>::operator[] creates the element if it does not exist

evgeniy-brazhnikov:
Last commit has been verified via jenkins-build.sh on Dev VM. Finished successfully

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Agree. The message is misleading. Will fix.
Will add number of bytes been read below

server/s3_copy_object_action.cc
amitwac2608: Yes, but keep note of this optimization... Address this in our optimization story.

amitwac2608 (approved):

PR #437, opened 2020-12-08
Handle deletions in dev environment for delayed delete

AtitaShirwaikar:
## Problem Statement
_[EOS-14519](https://jts.seagate.com/browse/EOS-14519):_
_Handle deletions in dev environment using backgroundelete for delayed delete_

## Problem Description
Handle deletions in dev environment using backgroundelete
whenever S3 IO is performed using jenkins build job or
any manually executed S3 IO (outside of the jenkins session) by manually running all services (e.g, s3server, s3auth, haroxy, ldap, motr) there should be no object leak in the dev environment

## Solution Overview
_1) Ensure that rabbit mq is installed and is up and running_
_2) Stop/Start s3background producer and consumer services_
_3) In case of cleanup mode or after tests are run, leave s3backgroundproducer and s3backgroundconsumer in a disabled state_

## Unit Test Cases
_1)./jenkins with STs should start and finish cleanly (no objects and index entries in motr, no background service running)_

_2)background services should also start by default on dev-vm when we do start-s3 and dev-stop-s3_

_3)multiple times start-s3 and dev-stop-s3_


jenkins-build.sh
```diff
@@ -259,6 +261,18 @@ if [ $cleanup_only -eq 1 ]; then
     set -e
 fi
 
+is_producer_running=1
+$USE_SUDO systemctl is-active s3backgroundproducer 2>&1 > /dev/null || is_producer_running=0
+if [[ $is_producer_running -eq 1 ]]; then
```
abhilekh268, line 266: if we intend to stop background services at this point ,can we simply stop them without this conditional check.

PR #438, opened 2020-12-09
EOS-15482: s3server support for reload of config file

knrajnambiar76:
When s3server gets SIGHUP signal, it will load all
modifiable parameters from config file, currently it will
load log_level.


sachitanands (approved):

server/s3_option.cc
```diff
@@ -673,6 +673,53 @@ bool S3Option::load_all_sections(bool force_override_from_config = false) {
   return true;
 }
 
+bool S3Option::reload_modifiable_options() {
+  try {
+    YAML::Node root_node = YAML::LoadFile(option_file);
+    if (root_node.IsNull()) {
+      return false;  // File Not Found?
```
DPG17, line 680: Do we need to log a message here at error level?
I don't see the caller adding message to log file when YAML::LoadFile() fails here.

DPG17 (approved):

server/s3_option.cc
knrajnambiar76: Added

PR #443, opened 2020-12-10
EOS-15436:S3:Reducing the footprint of log files.

SeagateChaDeepak:
**S3 Server changes :**

> Changed Entering to Entry.
> Changed Exiting to Exit.
> Function names will be present in only Entry, Exit, Constructor
> and Destructor Log messages.
> Req ID changed to Req.
> Added Stripped Request id which will be used in place of
> request id.
> Received Request with uri shortened to Req uri.
> Memory pool for unit_size shortened to mempool for size.
> Additional print added to display request id the first time
> along with its stripped counterpart.

**Auth Server changes :**

> "Requested action is -" changed to "Action -".
> "ACL authorization request is validated successfully" changed to "ACL
> authorization success".
> "Sending HTTP Response code" changed to "HTTP Response".
> Added Stripped Request id.
> Removed thread and request id print.

Signed-off-by: Deepak Mahale <deepak.mahale@seagate.com>

### DIFF
--- a/auth/server/src/main/java/com/seagates3/authserver/AuthServerGetHandler.java
+++ b/auth/server/src/main/java/com/seagates3/authserver/AuthServerGetHandler.java
@@ -92,6 +92,15 @@ class AuthServerGetHandler {
       AuthServerConfig.setReqId(BinaryUtil.getAlphaNumericUUID());
     }
 
+    String stripped_request_id = AuthServerConfig.getReqId();
+    AuthServerConfig.setStripedReqId(stripped_request_id);
+    LOGGER.info("Generating Stripped ReqId");
+
+    if (stripped_request_id.length() > 12) {
+      AuthServerConfig.setStripedReqId(
+          stripped_request_id.substring(stripped_request_id.length() - 12));
+    }
+
     if (httpRequest.uri().startsWith("/static")) {
       Path staticFilePath =
           Paths.get(AuthServerConstants.RESOURCE_DIR, httpRequest.uri());
@@ -150,7 +159,7 @@ class AuthServerGetHandler {
           HttpVersion.HTTP_1_1, requestResponse.getResponseStatus(),
           Unpooled.wrappedBuffer(responseBody.getBytes("UTF-8")));
 
-      LOGGER.info("Sending HTTP Response code [" + response.status() + "]");
+      LOGGER.info("HTTP Response [" + response.status() + "]");
     }
     catch (UnsupportedEncodingException ex) {
       LOGGER.error("UTF-8 encoding is not supported.");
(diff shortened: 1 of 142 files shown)


auth/server/src/main/java/com/seagates3/authserver/AuthServerGetHandler.java
```diff
@@ -92,6 +92,15 @@ void run() {
       AuthServerConfig.setReqId(BinaryUtil.getAlphaNumericUUID());
     }
 
+    String stripped_request_id = AuthServerConfig.getReqId();
+    AuthServerConfig.setStripedReqId(stripped_request_id);
```
sachitanands, line 96: add this as part of else condition in below code..

auth/server/src/main/java/com/seagates3/authserver/AuthServerPostHandler.java
```diff
@@ -72,6 +72,15 @@ public void run() {
           AuthServerConfig.setReqId(BinaryUtil.getAlphaNumericUUID());
         }
 
+        String stripped_request_id = AuthServerConfig.getReqId();
+        AuthServerConfig.setStripedReqId(stripped_request_id);
```
sachitanands, line 76: add this as part of else condition in below code..

server/request_object.cc
```diff
@@ -126,7 +126,11 @@ RequestObject::RequestObject(
 
   S3Uuid uuid;
   request_id = uuid.get_string_uuid();
-  s3_log(S3_LOG_DEBUG, request_id, "Constructor\n");
+  // last 12 characters
+  stripped_request_id = std::string(request_id.end() - 12, request_id.end());
```
amitwac2608, line 130: write a unit test to test that the stripped_request_id will always have 12 chars.

server/request_object.cc
```diff
@@ -216,7 +220,7 @@ void RequestObject::initialise() {
 }
 
 RequestObject::~RequestObject() {
-  s3_log(S3_LOG_DEBUG, request_id, "Destructor\n");
+  s3_log(S3_LOG_DEBUG, request_id, "%s Dtor\n", __func__);
```
amitwac2608, line 223: why to log "Dtor" !! __func__ would log the name of the function, and would be understood from this only that this is destructor !!

auth/server/src/main/java/com/seagates3/authserver/AuthServerPostHandler.java
SeagateChaDeepak: This part is required to print the entire request Id the first time its encountered.
The second setstripedreqid is used to set the actual stripped request id.

auth/server/src/main/java/com/seagates3/authserver/AuthServerGetHandler.java
SeagateChaDeepak: This part is required to print the entire request Id the first time its encountered.
The second setstripedreqid is used to set the actual stripped request id.

server/request_object.cc
SeagateChaDeepak: Done.

server/request_object.cc
SeagateChaDeepak: Done

sachitanands (approved):
AuthServer changes looks good.

knrajnambiar76 (approved):

PR #445, opened 2020-12-14
EOS-15439:S3: remove motr rpm dependency from s3 rpm

sachitanands:
Remove motr dependency from s3 rpm build process.


sachitanands:
pushed another changes hence these change are not required

PR #460, opened 2020-12-18
EOS-15555: copy object optimization

evgeniy-brazhnikov:

### DIFF
--- a/server/s3_copy_object_action.cc
+++ b/server/s3_copy_object_action.cc
@@ -259,23 +259,10 @@ void S3CopyObjectAction::copy_object() {
 void S3CopyObjectAction::read_data_block() {
   s3_log(S3_LOG_INFO, stripped_request_id, "%s Entry\n", __func__);
 
-  if (read_in_progress) {
-    s3_log(S3_LOG_DEBUG, stripped_request_id,
-           "%s Previous reading is in progress. Exit\n", __func__);
-    return;
-  }
-  if (!data_blocks_read.empty()) {
-    s3_log(
-        S3_LOG_DEBUG, stripped_request_id,
-        "%s Data from previous read has not been taken for writing yet. Exit\n",
-        __func__);
-    return;
-  }
-  if (bytes_left_to_read == 0) {
-    s3_log(S3_LOG_DEBUG, stripped_request_id,
-           "%s All data has been already read. Exit\n", __func__);
-    return;
-  }
+  assert(!read_in_progress);
+  assert(data_blocks_read.empty());
+  assert(bytes_left_to_read > 0);
+
   const auto n_blocks = std::min<size_t>(
       S3Option::get_instance()->get_motr_units_per_request(),
       (bytes_left_to_read + motr_unit_size - 1) / motr_unit_size);
@@ -316,6 +303,8 @@ void S3CopyObjectAction::read_data_block_success() {
     return;
   }
   if (copy_failed) {
+    assert(!write_in_progress);
+
     send_response_to_s3_client();
     return;
   }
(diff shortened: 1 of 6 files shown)


evgeniy-brazhnikov:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/299/

evgeniy-brazhnikov:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/303/

knrajnambiar76 (commented):
Please add comments in code

server/s3_copy_object_action.cc
```diff
@@ -292,13 +303,76 @@ void S3CopyObjectAction::read_data_block_success() {
...
+
+  for (size_t i = 0, n = data_blocks_read.size() - 1; i < n; ++i) {
+    const auto motr_block_size = data_blocks_read[i].second;
+    assert(motr_block_size == motr_unit_size);
+
+    // We can use multiplication, but something may change in the future
+    bytes_in_chunk_count += motr_block_size;
+  }
+  if (bytes_in_chunk_count >= bytes_left_to_read) {
+    s3_log(S3_LOG_ERROR, stripped_request_id, "Too many data has been read");
```
amitwac2608, line 338: Too **much**

amitwac2608: I am assuming this situation should not have happened. And we are in this if block because something is not good, and instead of discarding the extra read bytes, its good to set error here.

Is this understanding correct ?

server/s3_copy_object_action.cc
```diff
@@ -292,13 +303,76 @@ void S3CopyObjectAction::read_data_block_success() {
...
+    bytes_left_to_read = 0;
+  } else {
+    bytes_left_to_read -= r_last_block_size;
+  }
+  bytes_in_chunk_count += r_last_block_size;
+
+  s3_log(S3_LOG_DEBUG, request_id, "Got %zu bytes in %zu blocks",
+         bytes_in_chunk_count, data_blocks_read.size());
+
   if (!write_in_progress) {
```
amitwac2608, line 364: in case write_in_progress is "True", then should we log that some other write is in progress, so will write the chunk read above later. or something like this ?

This will help in debugging I guess, what do you say ?

ut/mock_s3_motr_reader.h
```diff
@@ -45,6 +45,7 @@ class MockS3MotrReader : public S3MotrReader {
                     std::function<void(void)> on_failed));
   MOCK_METHOD1(get_first_block, size_t(char** data));
   MOCK_METHOD1(get_next_block, size_t(char** data));
+  MOCK_METHOD0(extract_blocks_read, S3BufferSequence());
```
amitwac2608, line 48: MOCK_METHODN is deprecated.

Consider using below:
 MOCK_METHOD(S3BufferSequence, extract_blocks_read, ());

knrajnambiar76 (approved):

server/s3_copy_object_action.cc
knrajnambiar76, line 338: debug message makes more send if we log value of bytes_in_chunk_count and bytes_left_to_read

knrajnambiar76 (approved):

server/s3_copy_object_action.cc
evgeniy-brazhnikov: I thought we should detect such issuers. We can face with bugs in S3MotrReader class or Motr code.

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Agree

ut/mock_s3_motr_reader.h
evgeniy-brazhnikov: Sometimes ago someone (from leads) contradicted against destructing of uniform without reason.

server/s3_copy_object_action.cc
evgeniy-brazhnikov: Agree. Will do

amitwac2608: yes, I totally agree... But I wanted to understand is this situation ("too much read") is because of some corruption/problem in the system/stack or can happen commonly ?

ut/mock_s3_motr_reader.h
amitwac2608: "Deprecation of a method" is a good reason to go away with the "uniformity", right ?

amitwac2608 (approved):
Approving as my comments are minor, can be ignored or taken up in later PRs.

evgeniy-brazhnikov:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/309/

PR #581, opened 2020-12-25
[EOS-15721] Add dns round robin config generator script

dsurnin:


scripts/dns_rr_configs.py
```diff
@@ -0,0 +1,257 @@
...
+                5s )     ; minimum TTL
+;
+;
+;Name Server Information
+@               IN      NS      dns1.{zone}.
+dns1            IN      A       127.0.0.1
+
+;
+; A Records:
+; Configuring IP addresses should added here
```
t7ko-seagate, line 209: ... should be added ...

scripts/dns_rr_configs.py
```diff
@@ -0,0 +1,257 @@
...
+To install BIND use following command
+        #> yum install bind bind-utils
+
+To run BIND use following commands
+        #> systemctl enable named.service
+        #> systemctl start named.service
+
+To restart BIND after config update use following command
+        #> systemctl restart named.service
+        """)
```
t7ko-seagate, line 75: Either add here to "help", or create separate readme file which explains all needed steps in right order -- e.g.

yum install -y bind bind-utils
systemctl stop named.service # in case it is already started
./script.py -i ip1 ip2 ...
systemctl enable named.service
systemctl start named.service

Goal -- there must be a step-by-step manual, which can be just executed without prior knowledge of internals.

t7ko-seagate (changes requested):

scripts/dns_rr_configs.py
```diff
@@ -0,0 +1,257 @@
+#!/usr/bin/env python3
```
t7ko-seagate, line 1: Also please add JIRA ID in PR header.

t7ko-seagate (changes requested):

scripts/dns_rr_configs.py
dsurnin: done

scripts/dns_rr_configs.py
dsurnin: done

scripts/dns_rr_configs.py
dsurnin: removed BIND management commands from help

step-by-step manual is already written and could be found here
docs/LB_and_HA/dns_round_robin_config.md

t7ko-seagate (approved):

scripts/dns_rr_configs.py
```diff
@@ -0,0 +1,272 @@
...
+        description="""Configure BIND dns server to round robin IPs.""",
+        epilog=f"""Examples:
+To configure with default params: zone seagate.com and services s3, iam
+        #> {sys.argv[0]} --ip 192.168.1.0 192.168.1.1
+
+To configure for different zones and services
+        #> {sys.argv[0]} --ip 192.168.1.0 192.168.1.1 --zone test.zone --services srv1 srv2
+        """)
+
+    argz = parser.add_argument("-z", "--zone", type=str, required=False, default="seagate.com",
```
amitwac2608, line 61: should we not use "seagate" for the default zone name ?

I am thinking if this script runs in customer environment (correct me if I am wrong), then their zone should not be set to "seagate.com" (even if its harmless), but to some more generic name, like myzone.com.

amitwac2608 (commented):
minor comment.

scripts/dns_rr_configs.py
t7ko-seagate: In this case I propose we don't have any default at all and make this option a mandatory argument?  @amitwac2608 @dsurnin

dsurnin: this script configures local dns server and configures machine it was run on to use this server.
script does not provide any production ready server configuration just a fast and simple test environment.
it should not be recommended for end-user usage.

amitwac2608: Ok, we can go ahead then and merge this.

amitwac2608 (approved):

PR #614, opened 2021-01-12
EOS-16099 : Add Codacy token as an argument to the script for code coverage

pujamudaliar:
Add codacy token as an argument to the coverage script.

### DIFF
--- a/scripts/coverage/run_all_coverage.sh
+++ b/scripts/coverage/run_all_coverage.sh
@@ -21,12 +21,18 @@
 set -e
 set +o posix
 
+#Check if codacy token as an argument is passed
+if [[ $# -eq 0 ]] ; then
+    echo 'Error: No arguments supplied,please pass codacy token as an argument to the script'
+    exit 1
+fi
+
 # variables
 SCRIPT_PATH=$(readlink -f "$0")
 SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
 SRC_DIR="$(dirname "$(dirname "$SCRIPT_DIR" )")"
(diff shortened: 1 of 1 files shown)


scripts/coverage/run_all_coverage.sh
```diff
@@ -21,12 +21,18 @@
...
+    exit 1
+fi
+
 # variables
 SCRIPT_PATH=$(readlink -f "$0")
 SCRIPT_DIR=$(dirname "$SCRIPT_PATH")
 SRC_DIR="$(dirname "$(dirname "$SCRIPT_DIR" )")"
 # Default codacy project token for cortx-s3server
-CODACY_PROJECT_TOKEN=7a64ef82e1a049ab898fe838f1320594
+CODACY_PROJECT_TOKEN=$1
```
Shreya-18, line 35: There is already a way "-t" in getopt to accept `CODACY_PROJECT_TOKEN`. Couldn't we use that instead (make it mandatory option) and here set this variable="" ?
I can help you with this.

else if we are using $1, then `shift 1;` is necessary ...for getopt to fetch further arguments

scripts/coverage/run_all_coverage.sh
```diff
@@ -21,12 +21,18 @@
 set -e
 set +o posix
 
+#Check if codacy token as an argument is passed
```
Shreya-18, line 24: Please could you update the `usage()`.

Shreya-18 (approved):
Looks good, thanks for the changes!!

sachitanands (approved):

nileshgovande (approved):

PR #616, opened 2021-01-13
EOS:16252 Changed debug log level for s3backgrounddelete in build-rpm.sh

AtitaShirwaikar:
In s3 rpm build script, the default debug log level in s3_background_delete_config.yaml is taken.

While building the RPM , this level needs to be changed to DEBUG for s3backgrounddelete if ENABLE_DEBUG_LOG is set

### DIFF
--- a/rpms/s3/buildrpm.sh
+++ b/rpms/s3/buildrpm.sh
@@ -89,6 +89,7 @@ if ! [ -z "${GIT_VER}" ]; then
     if [ $ENABLE_DEBUG_LOG == 1 ]; then
         sed -i 's/#logLevel=DEBUG.*$/logLevel=DEBUG/g' auth/resources/authserver.properties.sample
         sed -i 's/S3_LOG_MODE:.*$/S3_LOG_MODE: DEBUG/g' s3config.release.yaml.sample
+        sed -i 's/file_log_level:.*$/file_log_level: 10/g' s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
     fi
     # For sake of test, attempt checkout of version
     git checkout ${GIT_VER}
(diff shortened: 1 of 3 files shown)


rpms/s3/buildrpm.sh
```diff
@@ -89,6 +89,7 @@ if ! [ -z "${GIT_VER}" ]; then
     if [ $ENABLE_DEBUG_LOG == 1 ]; then
         sed -i 's/#logLevel=DEBUG.*$/logLevel=DEBUG/g' auth/resources/authserver.properties.sample
         sed -i 's/S3_LOG_MODE:.*$/S3_LOG_MODE: DEBUG/g' s3config.release.yaml.sample
+        sed -i 's/file_log_level:.*$/file_log_level: 10/g' s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
```
sachitanands, line 92: current default value in config file is already set to 10..can we update default value a bit less in config file itself ?
when we set  "ENABLE_DEBUG_LOG " flag true then we can go for log_level as 10

AtitaShirwaikar: @sachitanands , your view is right, but debug is the lowest value.Refer to the table below

Level | Numeric value
-- | --
CRITICAL | 50
ERROR | 40
WARNING | 30
INFO | 20
DEBUG | 10
NOTSET | 0

sachitanands: can we set default bgdelete log level to "INFO" level and once we enable this flag we can set it to DEBUG simillar to Authserver.

nileshgovande: Agree to @sachinnerkar-art

nileshgovande (approved):

rpms/s3/buildrpm.sh
AtitaShirwaikar: Changed in 731206dc1fd81dd1f8f3f4e09c6f314599d1b684

sachitanands (approved):
Looks good

PR #623, opened 2021-01-18
Remove commit-id option from code coverage upload

pujamudaliar:
Commit-id option is not required to pass to code coverage script as "bash <(curl -Ls $codacy_coverage_url) report" uses latest commit id to upload the coverage report unless we need to explicitly upload against a specific commit-id. But if not then it is not required to pass.At present we are using latest commit-ids with default branch as main to upload the reports.

### DIFF
--- a/scripts/coverage/run_all_coverage.sh
+++ b/scripts/coverage/run_all_coverage.sh
@@ -54,9 +54,6 @@ run_coverage() {
 upload_report() {
     export CODACY_PROJECT_TOKEN
     export CODACY_API_BASE_URL=https://api.codacy.com
-
-    commit_id=$(git rev-parse $BRANCH)
-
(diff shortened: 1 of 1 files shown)


nileshgovande:
What is the harm if we keep it as-is ? @pujamudaliar

pujamudaliar:
> 
> 
> What is the harm if we keep it as-is ? @pujamudaliar

@nileshgovande : Within the script there is a command "git rev-parse main" which fails with the error message as "fatal: ambiguous argument 'main': unknown revision or path not in the working tree" because it runs inside the workspace (/root/jenkins_workspace/workspace/TEST-S3-Code-Coverage/cortx-s3server) created by Jenkins job.

nileshgovande (approved):

sachitanands (approved):

Shreya-18 (approved):

PR #635, opened 2021-01-21
EOS-16218:S3: Changed the config file of background delete to use message_bus framework

vimalraghwani20:
Changed the config file of background delete to use message_bus framework


SeagateChaDeepak (approved):
+1

PR #643, opened 2021-01-25
[EOS-16936] Integrate extended fakes into build/debug infra

dsurnin:


dsurnin:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-7.8.2003/466/

dev-starts3.sh
```diff
@@ -220,7 +220,8 @@ do
   pid_filename='/var/run/s3server.'$s3port'.pid'
   $valgrind_cmd s3server --s3pidfile $pid_filename \
            --motrlocal $local_ep:${motr_local_port} --motrha $ha_ep \
-           --s3port $s3port --fault_injection true $fake_params --loading_indicators --getoid true
+           --s3port $s3port --fault_injection true $fake_params --loading_indicators --getoid true \
+           --addb
```
t7ko-seagate, line 224: Pls confirm this is not a misprint.  If this is intended, please confirm we won't have problems with performance and disk usage in dev env with ADDB enabled.

server/motr_common.cc
```diff
@@ -168,6 +168,15 @@ void teardown_motr_op(struct m0_op *op) { teardown_motr_cancel_wait_op(op, 0); }
 void teardown_motr_cancel_wait_op(struct m0_op *op, int sec) {
```
t7ko-seagate, line 168: At some point, I saw some code in S3 which was doing the same teardown directly, not using this function.  Pls search for op_fini/op_free in code and make sure we won't have memory leak or crash.

t7ko-seagate (changes requested):

server/motr_common.cc
dsurnin: those functions are now called from teardown only

dev-starts3.sh
dsurnin: release configuration has addb enabled by default, I think it would be good to debug/develop on configuration as close to release as possible
addb has low performance impact and we usually dev-starts3 script is used not for performance tests but tests on dev VM
all addb data are removed on each jenkins-build script run or could be forced by jenkins-build --cleanup_only option, so I do not think there will be any significant extra disk usage.

t7ko-seagate (approved):

server/motr_common.cc
```diff
@@ -168,6 +168,15 @@ void teardown_motr_op(struct m0_op *op) { teardown_motr_cancel_wait_op(op, 0); }
...
   if (!op) return;
 
+  // Free fake m0_op for read in case of faked create and open
+  // NOTE: during operation init m0_op should be allocated accordingly
+  // see s3_motr_wrapper.cc::ConcreteMotrAPI::motr_obj_op
+  if (S3Option::get_instance()->is_fake_motr_obj_op_read(
+          (m0_obj_opcode)op->op_code)) {
+    free(op);
```
amitwac2608, line 176: do we need to assign nullptr to 'op' after 'free' call ?

amitwac2608 (commented):
minor comment, looks good

knrajnambiar76 (approved):

server/motr_common.cc
dsurnin: we can, but it is not necessary since this func is in the middle of the parent object cleanup logic

PR #669, opened 2021-02-04
[EOS-16014] Remove 'create_timestamp' attribute in the JSON value in Global bucke…

pranavPawar07:
…t list index (Revert of recovery feature)
Signed-off-by: pranav <pranav.pawar@seagate.com>


knrajnambiar76 (approved):
Have commit message  mentioning revert of recovery changes, so that if needed we can search later

nileshgovande (approved):

PR #697, opened 2021-02-16
EOS-17265: Initiate multipart upload as per new design

knrajnambiar76:
1) Removal of motr oid creation
2) Appending upload id to object name and some other changes

### DIFF
--- a/server/s3_object_metadata.cc
+++ b/server/s3_object_metadata.cc
@@ -99,7 +99,6 @@ void S3ObjectMetadata::initialize() {
   if (is_multipart) {
     index_name = get_multipart_index_name();
     system_defined_attribute["Upload-ID"] = upload_id;
-    (void)system_defined_attribute["Part-One-Size"];
   } else {
     index_name = get_object_list_index_name();
   }
@@ -129,6 +128,11 @@ S3ObjectMetadata::S3ObjectMetadata(
     object_name = request->get_object_name();
   } else {
     object_name = std::move(objectname);
+  }
+  if (upload_id.size() != 0) {
+    // for multipart uoload object name is appended with
+    // upload id
+    object_name += "|" + upload_id;
   }
   if (motr_api) {
     s3_motr_api = std::move(motr_api);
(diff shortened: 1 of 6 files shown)


server/s3_post_multipartobject_action.cc
```diff
@@ -311,118 +240,16 @@ void S3PostMultipartObjectAction::check_object_state() {
...
-    return;
-  }
-
-  if (motr_writer->get_state() == S3MotrWiterOpState::exists) {
-    collision_occured();
-  } else if (motr_writer->get_state() == S3MotrWiterOpState::failed_to_launch) {
-    s3_log(S3_LOG_WARN, request_id, "Create object failed.\n");
-    set_s3_error("ServiceUnavailable");
-    send_response_to_s3_client();
   } else {
```
amitwac2608, line 243: indentation seems to be wrong..

amitwac2608 (approved):
minor comment about indentation

server/s3_object_metadata.cc
```diff
@@ -130,6 +129,11 @@ S3ObjectMetadata::S3ObjectMetadata(
   } else {
     object_name = std::move(objectname);
   }
+  if (upload_id.size() != 0) {
+    // for multipart uoload object name is appended with
```
DPG17, line 133: "uoload"...should be "upload" -

ut/s3_post_multipartobject_action_test.cc
```diff
@@ -153,223 +148,6 @@ TEST_F(S3PostMultipartObjectTest, VaidateEmptyTags) {
                action_under_test->get_s3_error_code().c_str());
 }
 
-TEST_F(S3PostMultipartObjectTest, UploadInProgress) {
```
DPG17, line 156: Can we have all these below TCs as is, but in a disabled state, say, by using #if 0 pragma. Later, when functionality is complete, we can re-enable them. You may also want to mark a label for such work in code, so that one can search for such term in the entire codebase and work on those items later. We may label it as "TODO:s3fault"

DPG17 (approved):

ut/s3_post_multipartobject_action_test.cc
knrajnambiar76: All Uts succeding with this changes, these Uts not needed as we have removed some functions

PR #748, opened 2021-03-08
EOS-18158/EOS-17244:Auth:limit max no of Account and IAM user in authserver

sachitanands:
Set max limit of Account, IAMUser creation and make it configurable into Authserver.properties.
**Scenario**
There are scenarios where 
1) mulit-thread create operation may result into exceeding account/user max creation limit specified in authserver.properties
2) mulitnode environment may process create() request simultaneously hence we need to control this behaviour.
 
**Fix** : 
After creation api call before sending response back to client we should check ldap entries (account/user) and check if account entries are exceeding to given limit then delete newly created account else send success response 


nileshgovande:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Mini-Provisioning/job/cortx-s3server/774/
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Component-Test/job/S3Server-PreMerge-Test/725/

sachitanands:
Review comments:
1) remove recovery account from authserver.properties
2) in fetchAll (List Accounts) we should add internal account count to maxldapsearch 
3) At the end of create operation we should delete newly created account if maxAccountlimit exceeds   (Same for IAM user)

auth/resources/authserver.properties.sample
```diff
@@ -26,6 +26,8 @@ httpsPort=28051
 defaultHost=0.0.0.0
 logFilePath=/var/log/seagate/auth/server
 samlMetadataFileName=saml-metadata.xml
+maxAccountLimit=1001
```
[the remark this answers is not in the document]
sachitanands: not required here since internal accounts need not to be mentioned here
with latest discussion we will keep this limit as per user specification and we will consider "s3InternalAccounts" list for calculating internal account  count.

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -99,8 +100,26 @@ public ServerResponse create() {
...
         LOGGER.info("Creating account: " + name);
+        int maxAccountLimit = AuthServerConfig.getMaxAccountLimit();
+        int internalAccountCount =
+            AuthServerConfig.getS3InternalAccounts().size();
+        int maxAllowedLdapResults = maxAccountLimit + internalAccountCount;
+
+        try {
+          accountCount = accountDao.getTotalCountOfAccounts();
+
+          if (accountCount >= maxAllowedLdapResults) {
```
abhilekh268, line 113: this check should be added in ListAccounts?....are we just throwing error if accounts exeed this amount without listing any account or after listing 1000 accounts

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -171,6 +190,22 @@ public ServerResponse create() {
           LOGGER.error("Create account delay failing - ", e);
           Thread.currentThread().interrupt();
         }
+        // Handle multi-threaded create API calls by deleting extra accounts,
+        // if account limit creation exceeded due to multiple thread/multiple
+        // node create() API calls.
+        try {
+          accountCount = accountDao.getTotalCountOfAccounts();
```
abhilekh268, line 197: it contains internal accounts also?

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -171,6 +190,22 @@ public ServerResponse create() {
...
+
+          if (accountCount > maxAllowedLdapResults) {
+            // delete newly created account since we exceeded account
+            // creation limit.
+            accountDao.ldap_delete_account(account);
```
abhilekh268, line 202: can't we check accounts present already and not create only if limit is 1000 already

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -171,6 +190,22 @@ public ServerResponse create() {
...
+            return accountResponseGenerator.internalServerError();
```
bkirunge7, line 203: This should return error message saying "Account limit exceeded"

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -171,6 +190,22 @@ public ServerResponse create() {
...
+          }
+        }
+        catch (DataAccessException ex) {
```
bkirunge7, line 206: can you catch exception separately for getTotalCountOfAccounts and ldap_delete_account to differentiate what failed

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: we already have this check for FetchAll() method for listing accounts and there we are fetching records till <maxlimit> + internal account count  and next records will be ignored.

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: yes. this will return total ldap entry count which contains all the accounts created including internal accounts.

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: that check is already there but we came across this scenario where 
lets say we have 999 accounts created in ldap and two parallel threads calls create() calls , in this case there are changes of two account creation .
hence to avoid that we added this extra check after account creation is done.

This scenario will occur because of concurrent request received by either multiple threads or multiple node/instance of authserver.

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: lets say we have 999 account and two threads parallelly try to do createAccount() .
1)  Since both thread starts same time they will add entries in ldap e.g. thread-1 adds 1000th account and thread-2 adds 1001th account.
2) at this point threads will check total ldap entry count and both receives value greater than 1000 hence they will delete their newly created accounts i.e. 1000th and 1001th . 
3) in case of successful deletion we will still have 999 entries in ldap ..
Hence we  should not send "Account limit exceed" message since both newly created account got deleted.
so that next time client can send another request which will create 1000th account.

shalakadharap (approved):

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: done

auth/resources/authserver.properties.sample
```diff
@@ -26,14 +26,18 @@ httpsPort=28051
 defaultHost=0.0.0.0
 logFilePath=/var/log/seagate/auth/server
 samlMetadataFileName=saml-metadata.xml
+# Max limit of entries for account, iamuser and ldapsearch
+# operations can be configured using below three paramters.
+maxAccountLimit=1000
```
AjinkyaDhumal, line 31: Is 1000 ok or should it be 1001 ?

sachitanands: yes. t will be 1000 since this is client configuration parameters which will represent client account limit only . 
For internal accounts we are using another method : AuthServerConfig.getS3InternalAccounts().size(); 
this method will return internal account count which we are considering while checking ldap entries

sachitanands:
Pre-merge : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Component-Test/job/S3Server-PreMerge-Test/767/consol 
(In-Progress)

auth/server/src/main/java/com/seagates3/controller/AccountController.java
bkirunge7: what's the error message sent to client in this case

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -172,6 +191,30 @@ public ServerResponse create() {
...
+        try {
+          if (accountCount > maxAllowedLdapResults) {
+            // delete newly created account since we exceeded account
+            // creation limit.
+            accountDao.ldap_delete_account(account);
+            return accountResponseGenerator.internalServerError();
+          }
+        }
+        catch (DataAccessException ex) {
+          return accountResponseGenerator.internalServerError();
```
bkirunge7, line 215: log error message with more details

auth/server/src/main/java/com/seagates3/controller/UserController.java
```diff
@@ -56,45 +58,91 @@ public UserController(Requestor requestor,
...
-                     ":user/" + user.getName();
-        LOGGER.debug("Creating and setting ARN - " + arn);
-        user.setArn(arn);
-        try {
-            userDAO.save(user);
-        } catch (DataAccessException ex) {
-            return userResponseGenerator.internalServerError();
+      }
+      catch (DataAccessException ex) {
+        return userResponseGenerator.internalServerError();
```
bkirunge7, line 84: log error message with details

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: it will be InternalServer Error

auth/server/src/main/java/com/seagates3/controller/UserController.java
sachitanands: done

auth/server/src/main/java/com/seagates3/controller/AccountController.java
sachitanands: done

sachitanands: done

auth/server/src/main/java/com/seagates3/controller/UserController.java
sachitanands: done

sachitanands:
Pre-merge : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Component-Test/job/S3Server-PreMerge-Test/788/

AjinkyaDhumal (approved):

sachitanands:
Pre-merge job passed : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Component-Test/job/S3Server-PreMerge-Test/790/

sachitanands:
Mini-prov job : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Mini-Provisioning/job/cortx-s3server/854/ (In-progress)

PR #756, opened 2021-03-11
EOS-18564:S3 Added reset and prepare phase for mini provisoner

vimalraghwani20:
- Added reset phase in the s3setup
- Added prepare phase in s3setup
- Added prepare phase in the  _s3setup


scripts/provisioning/_s3_setup
```diff
@@ -80,6 +80,11 @@ def main():
       ResetCmd(args.config).process()
       sys.stdout.write('INFO: reset successful.\n')
 
+    elif args.command == 'prepare':
```
amitwac2608, line 83: you have not added the prepare command to the arg parser above !!!

vimalraghwani20: fixed.

scripts/provisioning/_s3_setup
```diff
@@ -49,6 +49,9 @@ def main():
   testCmd = subparsers.add_parser('reset', help='setup reset')
   testCmd.add_argument("--config", help='config URL', type=str, required=True)
 
+  testCmd = subparsers.add_parser('prepare', help='setup prepare')
```
amitwac2608, line 52: rename it as 'prepareCmd'.

scripts/provisioning/_s3_setup
```diff
@@ -49,6 +49,9 @@ def main():
...
+  testCmd.add_argument("--config", help='config URL', type=str, required=True)
```
amitwac2608, line 53: I think lets make '--config' as optional, not mandatory, as we are really not using the option in prepare.

scripts/provisioning/preparecmd.py
```diff
@@ -0,0 +1,40 @@
...
+  def __init__(self, config: str):
+    """Constructor."""
+    try:
+      super(PrepareCmd, self).__init__(config)
+    except Exception as e:
+      raise e
+
+  def process(self):
+    """Main processing function."""
+    retval = 0
```
amitwac2608, line 38: we can just put 'pass', instead of returning any value, as no other command class is returning anything from "process"

scripts/provisioning/setup.yaml
```diff
@@ -36,6 +36,9 @@ s3:
   reset:
     cmd: /opt/seagate/cortx/s3/bin/s3_setup reset
     args: --config $URL
+  prepare:
+    cmd: /opt/seagate/cortx/s3/bin/s3_setup prepare
+    args: --config $URL
```
amitwac2608, line 41: I think we should put 'null' here as we do not use config option in prepare. Same we can do for post_install. Can you please check with Provisioner team (@nitin-seagate  or @ypise ) if we can use 'null' value ?

gowthamchinna:
run premerge-test

scripts/provisioning/preparecmd.py
vimalraghwani20: fixed.

scripts/provisioning/_s3_setup
vimalraghwani20: Lets keep this as it is as of now. as this is just prototype. 
Once we start implementing prepare phase, we can take the decision to make it optional or not.

amitwac2608 (approved):

scripts/provisioning/s3_setup
```diff
@@ -41,6 +41,10 @@ test                     run s3 sanity tests
 
 cleanup                  cleanup S3 account created for backgrounddelete service
 
+reset                    delete all logs related to S3
```
nileshgovande, line 44: all logs, metadata etc. (which was populated as a result of the test phase)

nileshgovande (approved):

PR #766, opened 2021-03-15
EOS-18618 - Fix memory leaks from evhttp_uridecode and replace deprecated evhttp_decode_uri

shahidSTX:
EOS-18618 - Fix memory leaks from evhttp_uridecode and replace deprecated evhttp_decode_uri

Files modified :
   server/s3_copy_object_action.cc
   server/s3_post_multipartobject_action.cc
   server/s3_put_chunk_upload_object_action.cc
   server/s3_put_object_action.cc

Testing :
On CLI :
`put-object pbucket1/object - result(PASS)`
`copy-object pbucket1/object - result(PASS)`
End result : put-object and copy-object via CLI is successful for regular objects.

Signed-off-by: Mohammed Shahid <mohammed.shahid@seagate.com>

### DIFF
--- a/server/s3_copy_object_action.cc
+++ b/server/s3_copy_object_action.cc
@@ -91,7 +91,10 @@ void S3CopyObjectAction::get_source_bucket_and_object() {
       source_object_name = source.substr(separator_pos + 1);
     }
   }
-  source_object_name = evhttp_decode_uri(source_object_name.c_str());
+  char* decode_uri = evhttp_uridecode(source_object_name.c_str(), 0, NULL);
+  source_object_name = decode_uri;
+  free(decode_uri);
+  decode_uri = NULL;
   s3_log(S3_LOG_DEBUG, "", "%s Exit", __func__);
 }
 
--- a/server/s3_post_multipartobject_action.cc
+++ b/server/s3_post_multipartobject_action.cc
@@ -171,7 +171,7 @@ void S3PostMultipartObjectAction::parse_x_amz_tagging_header(
     for (struct evkeyval* header = key_value.tqh_first; header;
          header = header->next.tqe_next) {
 
-      decoded_key = evhttp_decode_uri(header->key);
+      decoded_key = evhttp_uridecode(header->key, 0, NULL);
       s3_log(S3_LOG_DEBUG, request_id,
              "Successfully parsed the Key Values=%s %s", decoded_key,
              header->value);
--- a/server/s3_post_multipartobject_action.cc
+++ b/server/s3_post_multipartobject_action.cc
@@ -176,6 +176,8 @@ void S3PostMultipartObjectAction::parse_x_amz_tagging_header(
              "Successfully parsed the Key Values=%s %s", decoded_key,
              header->value);
       new_object_tags_map[decoded_key] = header->value;
+      free(decoded_key);
+      decoded_key = NULL;
     }
     validate_tags();
   } else {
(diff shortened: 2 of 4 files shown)


server/s3_post_multipartobject_action.cc
```diff
@@ -171,7 +171,7 @@ void S3PostMultipartObjectAction::parse_x_amz_tagging_header(
     for (struct evkeyval* header = key_value.tqh_first; header;
          header = header->next.tqe_next) {
 
-      decoded_key = evhttp_decode_uri(header->key);
+      decoded_key = evhttp_uridecode(header->key, 0, NULL);
```
evgeniy-brazhnikov, line 174: "The returned string must be freed by the caller."
Did you forget free() ?

shahidSTX: No, decoded_key is allocated memory from heap and is used by :
`new_object_tags_map[decoded_key] = header->value;`
This new_object_tags_map is used by validate_object_xml_tags and lives throughout the duration of the program, so we cannot free it. I tried freeing it which was causing failures.

evgeniy-brazhnikov: It's strange. A compiler must create temporary std::string object for the expression new_object_tags_map[decoded_key].
And decoded_key variable must be useless just after the expression.
Can you please retest

shahidSTX: new_object_tags_map is a std::map, not a temporary string, and decoded_key is one of its mapped entry.
I did testing couple of times and saw failures every time I tried to free decoded_key.

shahidSTX: Done, freeing up after the expression works.

server/s3_copy_object_action.cc
```diff
@@ -91,7 +91,10 @@ void S3CopyObjectAction::get_source_bucket_and_object() {
       source_object_name = source.substr(separator_pos + 1);
     }
   }
-  source_object_name = evhttp_decode_uri(source_object_name.c_str());
+  char* decode_uri = evhttp_uridecode(source_object_name.c_str(), 0, NULL);
+  source_object_name = decode_uri;
+  free(decode_uri);
+  decode_uri = NULL;
```
evgeniy-brazhnikov, line 97: For future: it's redundant in such cases. The variable isn't used later. Moreover it get out of scope on the next line.

evgeniy-brazhnikov (approved):

amitwac2608 (approved):

server/s3_put_chunk_upload_object_action.cc
```diff
@@ -208,11 +208,13 @@ void S3PutChunkUploadObjectAction::parse_x_amz_tagging_header(
     for (struct evkeyval* header = key_value.tqh_first; header;
          header = header->next.tqe_next) {
 
-      decoded_key = evhttp_decode_uri(header->key);
+      decoded_key = evhttp_uridecode(header->key, 0, NULL);
```
bkirunge7, line 211: does this set decoded_key to null if evhttp_uriecode fails

shahidSTX: Yes.

server/s3_copy_object_action.cc
knrajnambiar76, line 97: Not needed to make it NULL as decode_uri is local

knrajnambiar76 (approved):

PR #775, opened 2021-03-18
EOS-17297: S3faulttolerence - Get Object (Simple) API support

DPG17:
Get object API changes


server/s3_get_object_action.cc
```diff
@@ -183,6 +258,154 @@ void S3GetObjectAction::set_total_blocks_to_read_from_object() {
   }
 }
 
+void S3GetObjectAction::set_total_blocks_to_read_from_fragmented_object() {
+  // to read complete object, total number blocks to read is equal to total
```
amitwac2608, line 262: I think the comment does not make sense !!! Something missed in the sentence ?

amitwac2608 (approved):
looks good

server/s3_get_object_action.cc
DPG17: ok. Will remove that comment.

server/s3_get_object_action.cc
```diff
@@ -145,17 +150,86 @@ void S3GetObjectAction::validate_object_info() {
...
+             total_blocks_in_object);
+    } else {
+      unsigned int ext_entry_index = 0;
+      // Object is fragmented, total objects is fragments + 1
+      total_objects = object_metadata->get_number_of_fragments() + 1;
+      s3_log(S3_LOG_DEBUG, request_id,
+             "Total objects (including primary): (%zu)\n", total_objects);
+      for (unsigned int i = 0; i < total_objects; i++) {
+        struct S3ExtendedObjectInfo obj_info;
+        if (i == 0) {
```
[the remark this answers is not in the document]
DPG17: First object is main object or so called primary object. It is the record with key as object_name (as regular object in object list index). Further objects are at key ~object_name|...  .  As all these are stored in a vector, 0th index corresponds to first/primary object.

server/s3_object_metadata.cc
```diff
@@ -733,6 +735,7 @@ int S3ObjectMetadata::from_json(std::string content) {
   pvid_str = newroot["PVID"].asString();
   obj_fragments = newroot["FNo"].asUInt();
   obj_parts = newroot["PRTS"].asUInt();
+  primary_obj_size = newroot["Size"].asUInt64();
```
knrajnambiar76, line 738: Check if this is a member then call newroot["Size"].asUInt64();
Cases when upgrade happens this member may not be there, which may lead to crash

server/s3_get_object_action.cc
DPG17: Updated the comment and set it at the correct code line

knrajnambiar76 (approved):

server/s3_object_metadata.cc
DPG17: Done. Add isMember() for "FNo", as this is the key field for the new metadata format

PR #782, opened 2021-03-22
[EOS-18885]: S3 R1 DI: Object protection

max-seagate:
The features that were done:
- checksums for normal PUT: was done already
- checksums for multipart PUT: for each part and for CompleteMultipartUpload was done already
- checksums for normal GET: we implemented it
- multipart upload: "merge" checksums: PUT case was done already, GET case: we implemented it
- disable ranged read: done by us
- save part size for each part of multipart upload: we added saving part size for each part except the last (S3server only supports same size of all parts except the last)
- config: enable checksum verification during GET: we added an option
- change UTs/STs: we added configuration options in the way that default behavior is the same, so existing UTs and STs pass. We've changed the options for provisioned systems to disable ranged read and enable checksum verification during write.
- add a way for QA to test the new code: introduced a configuration option that injects failures in the following way: write buffers that are sent to Motr are corrupted after checksumming, but before sending them to Motr if the configuration option S3_ST_MD5_ERROR_INJECT_ENABLED is set to true. This ensures that reads of the same object would have checksum mismatch.
- performance: checksums were enabled on PUT (both normal and multipart), so this behavior hasn't been changed. We've added checksums on read, so read performance has to be tested. We haven't done it on a hardware cluster.
- test on OVA: we've tested this on 2 VMs with builds from source. We haven't tested it on OVA.
- IEM alerts: we've added them.
- documentation: new options are documented.

To run the test:
```
time st/clitests/integrity.py --auto-test-all
```
`aws s3` and `aws s3api` must be configured and must work.


rebuildall.sh
```diff
@@ -307,10 +307,10 @@ if [ $just_gen_build_file -eq 1 ]; then
 fi
 
 
-if [ $no_check_code -eq 0 ]
-then
-  ./checkcodeformat.sh
-fi
+# if [ $no_check_code -eq 0 ]
```
knrajnambiar76, line 310: why this commented ?

runst.sh
```diff
@@ -0,0 +1,17 @@
+export LD_LIBRARY_PATH="$(pwd)/third_party/motr/motr/.libs/:"\
```
knrajnambiar76, line 1: whats the need of this script

server/s3_get_object_action.cc
```diff
@@ -494,6 +494,17 @@ void S3GetObjectAction::send_data_to_client() {
...
+    std::string s_md5_calc = motr_reader->get_content_md5_base64();
+    std::string s_md5_read = object_metadata->get_md5();
+    s3_log(S3_LOG_ERROR, request_id, "MD5 calculated: %s, MD5 read %s", /* XXX */
+           s_md5_calc.c_str(), s_md5_read.c_str());     /* XXX */
+    s3_log(S3_LOG_DEBUG, request_id, "MD5 calculated: %s, MD5 read %s",
+           s_md5_calc.c_str(), s_md5_read.c_str());
+
+    if (s_md5_calc != s_md5_read) {
+      s3_log(S3_LOG_ERROR, request_id, "Content MD5 mismatch\n");
+      set_s3_error("BadDigest");
```
knrajnambiar76: Better if error is set to "InternalError"
BadDigest error is used if s3 client has provided MD5 of the file and there is mismatch, here s3 client didn't provide any md5 of the file

server/s3_get_object_action.cc
```diff
@@ -311,8 +317,14 @@ void S3GetObjectAction::check_full_or_range_object_read() {
...
     s3_log(S3_LOG_DEBUG, request_id, "Range found(%s)\n",
            range_header_value.c_str());
-    if (validate_range_header_and_set_read_options(range_header_value)) {
-      next();
+
+    if (S3Option::get_instance()->get_s3_ranged_read_enabled()) {
+      if (validate_range_header_and_set_read_options(range_header_value)) {
+	next();
+      } else {
+	set_s3_error("InvalidRange");
```
knrajnambiar76, line 325: you can set error code to "OperationNotSupported", since its disabled in option file

server/s3_get_object_action.cc
```diff
@@ -494,6 +511,24 @@ void S3GetObjectAction::send_data_to_client() {
...
+      }
+      std::string s_md5_read = object_metadata->get_md5();
+      s3_log(S3_LOG_DEBUG, request_id, "MD5 calculated: %s, MD5 read %s",
+	     s_md5_calc.c_str(), s_md5_read.c_str());
+      if (s_md5_calc != s_md5_read) {
+	s3_iem(LOG_ERR, S3_IEM_CHECKSUM_MISMATCH,
+	       S3_IEM_CHECKSUM_MISMATCH_STR,
+	       S3_IEM_CHECKSUM_MISMATCH_JSON);
+	s3_log(S3_LOG_ERROR, request_id, "Content checksum mismatch\n");
+	set_s3_error("BadDigest");
```
knrajnambiar76, line 529: set error to InternalError

server/s3_motr_reader.cc
```diff
@@ -344,6 +344,59 @@ size_t S3MotrReader::get_next_block(char **data) {
...
+    length = total_size_to_read - total_size_read;
+    s3_log(S3_LOG_DEBUG, "", "%s length changed after %zu: ", __func__, length);
+  }
+  if (multipart_part_size == 0) {
+    // non-multipart-upload case. Just calculate md5 of the entire object.
+    md5crypt.Update(*data, length);
+  } else {
+    // the object was created with multipart upload.
+    // Calculate md5 checksums for each part independently.
+    size_t begin = total_size_read;
```
knrajnambiar76, line 361: Not able to understand below logic, can you please put more comment here

server/s3_object_metadata.cc
```diff
@@ -605,8 +605,10 @@ std::string S3ObjectMetadata::to_json() {
   root["create_timestamp"] = current_time.get_isoformat_string();
 
   Json::FastWriter fastWriter;
-  return fastWriter.write(root);
-  ;
+  std::string s = fastWriter.write(root);
+  s3_log(S3_LOG_INFO, request_id,
```
knrajnambiar76, line 609: We are trying to reduce logging as much as possible, due to multiple s3servers... better if INFO mode logging is changes to DEBUG mode

server/s3_object_metadata.cc
```diff
@@ -656,6 +660,8 @@ std::string S3ObjectMetadata::version_entry_to_json() {
 int S3ObjectMetadata::from_json(std::string content) {
   s3_log(S3_LOG_DEBUG, request_id, "Called with content [%s]\n",
          content.c_str());
+  s3_log(S3_LOG_INFO, request_id, "S3ObjectMetadata::from_json(): s=%s\n",
```
knrajnambiar76, line 663: It can be changed to Debug mode

server/s3_part_metadata.cc
```diff
@@ -413,11 +413,16 @@ std::string S3PartMetadata::to_json() {
     root["User-Defined"][uit.first] = uit.second;
   }
   Json::FastWriter fastWriter;
-  return fastWriter.write(root);
+  std::string s = fastWriter.write(root);
+  s3_log(S3_LOG_INFO, request_id,
```
knrajnambiar76, line 417: DEBUG mode

rebuildall.sh
max-seagate: To simplify development iterations (or at least we thought so).

Restored.

runst.sh
max-seagate: Just a way to run STs. Got commited at some point.

Removed from this patch.

server/s3_get_object_action.cc
max-seagate: Done 4 lines below. This case is for ranged read enabled, but when the range validation fails.
Also fixed tabs & spaces (github doesn't work correctly with tabs, see EOS-11827 for the reference).

server/s3_get_object_action.cc
max-seagate: Done.

server/s3_get_object_action.cc
max-seagate: Done.

server/s3_object_metadata.cc
max-seagate: Removed completely. It was used to understand what kind of metadata is stored for each object, and it's no longer needed.

server/s3_object_metadata.cc
max-seagate: Removed.

server/s3_part_metadata.cc
max-seagate: Removed.

server/s3_motr_reader.cc
max-seagate: Re-implemented the chunk of code to make it easier to read, but haven't re-tested it. Will post another comment once this chunk of code is ready for review.

docs/object-protection-testing.md
```diff
@@ -0,0 +1,2 @@
+# Object Integrity testing
+
```
johnbent, line 2: Is this an empty file that was added?

s3config-test.yaml
```diff
@@ -19,6 +19,9 @@
 
 S3Config_Sections: [S3_SERVER_CONFIG, S3_AUTH_CONFIG, S3_MOTR_CONFIG, S3_THIRDPARTY_CONFIG]
 S3_SERVER_CONFIG:                                     # Section for S3 Server
+   S3_RANGED_READ_ENABLED: true                         # enable ranged reads
```
johnbent, line 22: Shouldn't the default be false since we are disabling this?

s3config-test.yaml
```diff
@@ -19,6 +19,9 @@
...
+   S3_READ_MD5_CHECK_ENABLED: false                     # enable data integrity checksum check on S3 GET
```
johnbent, line 23: Shouldn't the default be true?

s3config.yaml.sample
```diff
@@ -19,6 +19,9 @@
 
 S3Config_Sections: [S3_SERVER_CONFIG, S3_AUTH_CONFIG, S3_MOTR_CONFIG, S3_THIRDPARTY_CONFIG, S3_VERSION_CONFIG]
 S3_SERVER_CONFIG:
+   S3_RANGED_READ_ENABLED: true                         # enable ranged reads
+   S3_READ_MD5_CHECK_ENABLED: false                     # enable data integrity checksum check on S3 GET
```
johnbent, line 23: Shouldn't it be true here?

server/s3_get_object_action.cc
```diff
@@ -320,6 +331,13 @@ void S3GetObjectAction::read_object() {
   set_total_blocks_to_read_from_object();
   motr_reader = motr_reader_factory->create_motr_reader(
       request, object_metadata->get_oid(), object_metadata->get_layout_id());
+  size_t part_one_size = object_metadata->get_part_one_size();
+  if (part_one_size > 0) {
+    motr_reader->set_multipart_part_size(part_one_size);
+    std::string etag = object_metadata->get_md5();
+    std::string num_of_parts = etag.substr(etag.rfind('-') + 1);
```
johnbent, line 338: This looks like we have some naming convention embedded in the software.  If so, this worries me because it is easy to forget the convention and suddenly break code.  Perhaps there should be helper functions to build/parse this string so that the conventions can be isolated to just a single location?

server/s3_get_object_action.cc
```diff
@@ -475,6 +497,25 @@ void S3GetObjectAction::send_data_to_client() {
...
+        checksum_calculated = motr_reader->get_content_md5();
+      } else {
+        checksum_calculated = motr_reader->awsetag.finalize();
+      }
+      std::string checksum_read = object_metadata->get_md5();
+      s3_log(S3_LOG_DEBUG, request_id,
+             "checksum calculated: %s, checksum read %s",
+             checksum_calculated.c_str(), checksum_read.c_str());
+      if (checksum_calculated != checksum_read) {
+	s3_iem(LOG_ERR, S3_IEM_CHECKSUM_MISMATCH,
```
johnbent, line 512: On my screen, it looks like the indentation here is not correct.  Is this a hard tab vs soft tab issue maybe?

server/s3_motr_reader.cc
```diff
@@ -344,6 +344,39 @@ size_t S3MotrReader::get_next_block(char **data) {
...
+    // the object was created with multipart upload.
+    // Calculate md5 checksums for each part independently.
+    size_t remaining = length;
+    // Each iteration jumps to the next part or the end of the buffer,
+    // whichever comes first.
+    while (remaining > 0) {
+      // current absolute position in object for the purpose of checksumming
+      size_t pos = total_size_read + length - remaining;
+      size_t next_part_pos;
+      if (pos / multipart_part_size >= multipart_num_of_parts - 1)
```
johnbent, line 364: I don't know the S3 coding standard.  Brackets are not required for one-liners?  

Also, the logic in the if statement is not intuitive to me.  Might be a good place for a comment.

server/s3_motr_reader.cc
```diff
@@ -344,6 +344,39 @@ size_t S3MotrReader::get_next_block(char **data) {
...
+        next_part_pos = total_size_to_read;
+      else
+        next_part_pos = (pos / multipart_part_size + 1) * multipart_part_size;
```
johnbent, line 367: This logic also challenging.

server/s3_motr_reader.h
```diff
@@ -150,6 +152,15 @@ class S3MotrReader {
 
   S3MotrReaderOpState state;
 
+  // md5 for the content read from motr.
+  // In case of it's a GET for multipart upload it's calculated for each part
```
johnbent, line 156: Do we also do a checksum of checksums for the entire object or that's not necessary?

server/s3_motr_writer.cc
```diff
@@ -473,6 +474,30 @@ void S3MotrWiter::write_content() {
...
   set_up_motr_data_buffers(rw_ctx, data_items, motr_buf_count);
+
+  if (S3Option::get_instance()->get_s3_st_md5_error_inject_enabled()) {
+    struct m0_bufvec *bv = rw_ctx->data;
+    if (rw_ctx->ext->iv_index[0] == 0) {
+      char first_byte = *(char *)bv->ov_buf[0];
+      s3_log(S3_LOG_DEBUG, "", "%s first_byte=%d", __func__, first_byte);
+      switch (first_byte) {
+        case 'z':  // zero
+	  corrupt_fill_zero = true;
```
johnbent, line 485: Indentation looks off here.

server/s3_motr_writer.cc
```diff
@@ -473,6 +474,30 @@ void S3MotrWiter::write_content() {
...
+          // corrupt the first byte
+          *(char *)bv->ov_buf[0] = ~first_byte;
+          break;
+        case 'k':  // OK
+          break;
+      }
+    }
+    if (corrupt_fill_zero) {
+      for (uint32_t i = 0; i < bv->ov_vec.v_nr; ++i)
+        memset(bv->ov_buf[i], 0, bv->ov_vec.v_count[i]);
```
johnbent, line 497: Is this the function that injects errors?  Why are there two different ways to inject failures?  Seems unnecessarily complicated.  Just typecast the buffer to a char array and increment the first char.  Simple, easy, foolproof. No?

server/s3_motr_writer.h
```diff
@@ -119,6 +119,11 @@ class S3MotrWiter {
   // buffer currently used to write, will be freed on completion
   std::shared_ptr<S3AsyncBufferOptContainer> write_async_buffer;
 
+  // used for checksum calculation testing
+  // fill entire object with zeroes after checksum calculation, but before
+  // writing to Motr
+  bool corrupt_fill_zero = false;
```
johnbent, line 125: Presumably there is an easy way to set this without having to change and recompile the code, right?

johnbent (changes requested):
Looks good.  A few questions and a few places where I think comments would be useful and where the indentation looks off.

docs/object-protection-testing.md
max-seagate: Yes, work in progress.

server/s3_motr_writer.h
max-seagate: Right. Just enable `S3_ST_MD5_ERROR_INJECT_ENABLED` and pass an object with 'z' in the first byte of the object.

server/s3_motr_writer.cc
max-seagate: 2 kinds of corruption were added: change of the first char and zeroing the buffers to ensure that a) checksums are calculated for the entire object and b) change of each byte would be detected.

server/s3_motr_reader.h
max-seagate: Multipart upload is done in the following way: parts are uploaded one by one, then there is CompleteMultipartUpload call that is only about changing the metadata. To calculate the entire object md5 checksum we would have to read the entire object during CompleteMultipartUpload. There may be a hash that allows to calculate full hash using partial hashes, but md5 checksums were already implemented in the code for PUT and UploadPart, so sticking with md5 checksums was the easiest way to go.

s3config-test.yaml
max-seagate: It should. Let's see if S3server team prefers to keep it enabled in the branch or it could be disabled in the branch for all use cases.

s3config-test.yaml
max-seagate: Yes, I'd have to disable a number of UTs and STs for ranged read to make this work, and it would take more time. Let's see what is the preference of the team.

s3config.yaml.sample
max-seagate: Same as before.

server/s3_get_object_action.cc
max-seagate: Excellent point. This etag format is created in `server/s3_aws_etag`. I just have refactored this piece of code to that file, so operations with this format doesn't leave a single file. There is still knowledge that the etag embeds # of parts, but know it's a matter of leaky abstractions.

server/s3_motr_writer.cc
max-seagate: > Just typecast the buffer to a char array and increment the first char. 

This is also done a few lines earlier. Not increment, because increment of a char (that might be signed) is not something I'd like to have because of 6.3.1.3/3 in C99 ([n1256.pdf](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)):

> Otherwise, the new type is signed and the value cannot be represented in it; either the result is implementation-defined or an implementation-defined signal is raised.

I just checked the code that is currently being reviewed and it uses `~` which may also have issues. I changed it to `!`, which results in either 0 or 1, which should be fine for any type. Thanks for pointing out, it helped to find another issue.

server/s3_get_object_action.cc
max-seagate: It's a well-known GitHub issue EOS-11827.

S3server uses spaces. Fixed here and in the rest of the patch, all tabs were found with `git diff cortx-1.0 | grep -P '^\+\t'`.

server/s3_motr_writer.cc
max-seagate: Fixed.

server/s3_motr_reader.h
max-seagate: Added this explanation to the code.

server/s3_motr_reader.cc
max-seagate: Added a few comments. Please let me know if the logic is still not clear.

server/s3_motr_reader.cc
max-seagate: Commented.

max-seagate: > Brackets are not required for one-liners?

I just checked this file and some other file that I have open in editor - seems like they are. Fixed.

server/s3_motr_reader.cc
johnbent: I'm still not following.  Maybe if you created some temporary variables.  Like

foo = pos / multipart\_part\_size; // where foo is some intuitive variable name

bar = multipart\_num\_of\_parts - 1; // where bar is some intuitive variable name

then

if ( foo >= bar) {

.....

server/s3_aws_etag.cc
```diff
@@ -128,3 +128,8 @@ std::string S3AwsEtag::finalize() {
 }
 
 std::string S3AwsEtag::get_final_etag() { return final_etag; }
+
+int S3AwsEtag::get_num_of_parts(std::string etag)
+{
+    return std::stoull(etag.substr(etag.rfind('-') + 1));
```
johnbent, line 134: Looks like you moved this logic into a helper function.  Excellent.  Thank you.  Can you please also add a comment specifying what the expected string format is to help the reader understand the convention?

johnbent (changes requested):

server/s3_motr_reader.cc
max-seagate: Added more comments and changed `if` condition to make it even more easier to read. Please re-check.

server/s3_aws_etag.cc
max-seagate: Added documentation just before `S3AwsEtag` class.

docs/object-protection-testing.md
max-seagate: Added documentation.

server/s3_aws_etag.cc
johnbent: Thanks.  Resolving conversation.

server/s3_motr_reader.cc
```diff
@@ -344,6 +344,65 @@ size_t S3MotrReader::get_next_block(char **data) {
...
+      size_t pos = total_size_read + length - remaining;
+      size_t next_part_pos;
+      // Current implementation supports only equal sized parts, except the
+      // last part. This allows to store only a part size (for all parts except
+      // the last) and total number of parts. Part boundaries could be derived
+      // from these 2 values.
+      // The following 'if' gets absolute position in object of the next part
+      // boundary.
+      if (pos >= multipart_part_size * (multipart_num_of_parts - 1)) {
+	// pos is over the bounrady between last and last but one part. The
```
johnbent, line 381: misspelled the word boundary here.

server/s3_motr_reader.cc
johnbent: Much better.  Thank you.  @knrajnambiar76 , does this work for you?

server/s3_motr_reader.cc
max-seagate: Fixed.

ut/s3_get_object_action_test.cc
```diff
@@ -1202,3 +1202,57 @@ TEST_F(S3GetObjectActionTest, RangeHeaderContainsSpacesOnly) {
       range_value));
 }
 
+TEST_F(S3GetObjectActionTest, ReadObjectOfSizeEqualToUnitSizeMD5Fail) {
```
t7ko-seagate, line 1205: Need similar test which validates code that handles multipart objects.

t7ko-seagate (changes requested):

docs/design/object-protection.txt
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
```
amitwac2608, line 2: 2021

max-seagate: Fixed.

st/clitests/integrity.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+
+import json
+import uuid
+import random
+import argparse
+
+import plumbum
```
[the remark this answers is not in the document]
max-seagate: It depends on how this test would be executed. If it would be a part of CI then the answer is likely "yes", but it's not now.
If it's only for a manual execution then it's for you (or someone else from S3 server team) to decide.

Please let me know about the decision. Or I could spend more time to write whatever is present in that library already and use `subprocess` directly.

server/s3_motr_reader.h
```diff
@@ -177,6 +188,7 @@ class S3MotrReader {
   void clean_up_contexts();
 
  public:
+  S3AwsEtag awsetag;
```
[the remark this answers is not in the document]
max-seagate: Moved to private, added a function to access `awsetag.finalize()`.

docs/object-protection-testing.md
```diff
@@ -0,0 +1,54 @@
+# Object Integrity testing
```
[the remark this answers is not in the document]
johnbent: Actually most .md files don't have license block.  For example, [https://github.com/Seagate/cortx-s3server/blob/main/SUPPORT.md.](https://github.com/Seagate/cortx-s3server/blob/main/SUPPORT.md.)  In fact, I don't know any .md files that have license block.  Those are only needed in source code files but not documentation files.

server/s3_iem.h
```diff
@@ -110,4 +110,10 @@ extern S3Option* g_option_instance;
   "Metadata has accessibility issues. Contact Seagate Support."
 #define S3_IEM_METADATA_CORRUPTED_JSON ""
 
+#define S3_IEM_CHECKSUM_MISMATCH "0030060004"
```
max-seagate, line 113: [* suggestion *] Add bucket, object name, Motr object id to the IEM message.

max-seagate: Also expected and calculated checksum.

docs/object-protection-testing.md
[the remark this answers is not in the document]
t7ko-seagate: @johnbent 

Can we please have this information communication team-wide in an email on some other appropriate way?  otherwise we were instructed to have License block in every .md file and we're trying to follow this, since this was the requirement.

server/s3_get_object_action.cc
```diff
@@ -468,14 +515,16 @@ void S3GetObjectAction::send_data_to_client() {
...
   s3_timer.stop();
 
-  if (data_sent_to_client != requested_content_length) {
-    read_object_data();
-  } else {
-    const auto mss = s3_timer.elapsed_time_in_millisec();
-    LOG_PERF("get_object_send_data_ms", request_id.c_str(), mss);
-    s3_stats_timing("get_object_send_data", mss);
-
+  if (checksum_mismatch) {
```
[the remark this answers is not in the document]
max-seagate: Restored old logic.

server/s3_motr_reader.cc
```diff
@@ -344,6 +344,65 @@ size_t S3MotrReader::get_next_block(char **data) {
   *data = (char *)motr_rw_op_context->data->ov_buf[iteration_index];
   data_read = motr_rw_op_context->data->ov_vec.v_count[iteration_index];
   iteration_index++;
+
+  size_t length = data_read;
+  if (total_size_to_read < total_size_read + length) {
+    length = total_size_to_read - total_size_read;
```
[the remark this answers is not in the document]
max-seagate: Explained in a comment just above the `if`. Please check and let me know if something is still not clear.

server/s3_aws_etag.cc
```diff
@@ -128,3 +128,8 @@ std::string S3AwsEtag::finalize() {
 }
 
 std::string S3AwsEtag::get_final_etag() { return final_etag; }
+
+int S3AwsEtag::get_num_of_parts(const std::string& etag)
+{
+    return std::stoull(etag.substr(etag.rfind('-') + 1));
```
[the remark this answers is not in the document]
max-seagate: Properly handled error here and added an assert on a level just above this function.

docs/object-protection-testing.md
```diff
@@ -0,0 +1,54 @@
...
+Set `S3_ST_MD5_ERROR_INJECT_ENABLED: true` and then
+
+    aws s3 mb s3://test
+    time st/clitests/integrity.py --auto-test-all
+
+`aws s3` and `aws s3api` must be configured and must work.
+
+## Manually
+
+In `S3_ST_MD5_ERROR_INJECT_ENABLED: true` mode: if the first byte of the file
```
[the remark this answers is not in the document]
dsurnin: make fi similar to s3

server/s3_motr_writer.cc
```diff
@@ -473,6 +475,30 @@ void S3MotrWiter::write_content() {
   ctx->cbs[0].oop_stable = s3_motr_op_stable;
   ctx->cbs[0].oop_failed = s3_motr_op_failed;
   set_up_motr_data_buffers(rw_ctx, data_items, motr_buf_count);
+
+  if (S3Option::get_instance()->get_s3_st_md5_error_inject_enabled()) {
```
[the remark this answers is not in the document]
dsurnin: done

server/s3_iem.h
dsurnin: Extended with bucket name, object name, motr id, md5 calc, md5 read

dsurnin:
jenkins pre-merge job
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-cortx-1.0/35/

st/clitests/integrity.py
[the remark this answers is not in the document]
dsurnin: checks are added to jenkins_buildd

ut/s3_get_object_action_test.cc
dsurnin: multipart test added

dsurnin:
jenkins pre-merge
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-cortx-1.0/37/

dsurnin (approved):

server/s3_motr_reader.h
```diff
@@ -206,6 +226,20 @@ class S3MotrReader {
 
   virtual void set_last_index(size_t index) { last_index = index; }
 
+  // copy-paste from server/s3_motr_writer.h
```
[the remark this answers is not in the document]
johnbent: Why are we copy-pasting code?  That's poor software engineering that causes unnecessary code bloat and introduces high probability of future errors.  We should never copy-paste code.  If code from elsewhere is needed, it needs to be be copied to a common location and shared.

t7ko-seagate (approved):

johnbent (approved):

dsurnin:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-cortx-1.0/42/

dsurnin:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-cortx-1.0/53/

ujjwalpl (approved):

PR #809, opened 2021-04-01
[EOS-17282]Multipart listing changes as per new design

knrajnambiar76:
Multipart listing to not disply uploadid which gets in as key
in the multipart table, listing to show same objects with diffferent
upload ids in case of simultaneous multipart upload of same object

Signed-off-by: Rajesh <rajesh.nambiar@seagate.com>

### DIFF
--- a/server/s3_object_list_response.cc
+++ b/server/s3_object_list_response.cc
@@ -32,6 +32,7 @@ S3ObjectListResponse::S3ObjectListResponse(std::string encoding_type)
       request_marker_uploadid(""),
       max_keys(""),
       response_is_truncated(false),
+      chop_uploadid(false),
       next_marker_key(""),
       max_uploads(""),
       next_marker_uploadid(""),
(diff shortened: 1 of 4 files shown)


server/s3_object_list_response.h
```diff
@@ -85,6 +86,7 @@ class S3ObjectListResponse {
   void set_response_is_truncated(bool flag);
   void set_next_marker_key(std::string next, bool url_encode = true);
   void set_next_marker_uploadid(std::string next);
+  void chop_uploadid_from_key() { chop_uploadid = true; }
```
amitwac2608, line 89: can we make this function inline ?

server/s3_object_list_response.cc
```diff
@@ -295,9 +296,15 @@ std::string& S3ObjectListResponse::get_multiupload_xml() {
   }
 
   for (auto&& object : object_list) {
+    std::string object_name = object->get_object_name();
+    if (chop_uploadid) {
+      // In case of multipart listing, chop uploadid from key
+      std::size_t pos = object_name.find("|");
+      object_name = object_name.substr(0, pos);
```
amitwac2608, line 303: can 'pos' variable have a not valid value, like std::npos ?

amitwac2608: Also we can have "|" in object name, right ?

Do we need handling of this scenario as well ?

amitwac2608: # aws s3api list-objects --bucket mybk
{
            "LastModified": "2021-04-01T11:50:56.000Z", 
            "ETag": "\"8c9e4b10ec6e7576c2c427f7fb89b58a\"", 
            "StorageClass": "STANDARD", 
            **"Key": "my_o|bj",** 
            "Owner": {
                "DisplayName": "new-acc", 
                "ID": "da9a0c3acd48427fbf95f8d3dfe5be182f52ca5d7b44413cb74ec394036f8e69"
            }, 
            "Size": 108
        }

knrajnambiar76: if pos is std::npos then substr will show complete string, I have put in check to call substr based on pos value

Since object name can have | (as pr aws s3 its suppose to be avoided, but not restricted), hence changed find() to rfind()

DPG17 (approved):

amitwac2608 (approved):

PR #831, opened 2021-04-13
[GitHub#833] S3 logger robustness, performance improvement and reorg proposal

just-now:
- *code compiles but NOT TESTED*;
 - removed ASPRINTF! and unneeded allocations for EVERY! log write!!!
 - uses tls-approach.
 - change the semantics of s3_iem() so that it becomes similar to s3 log.
 - use templated verbosity select method to be able to handle const-expressions inside s3_log macro
   example: s3_log(cond ? S3_LOG_DEBUG : S3_LOG_ERROR, ...).


server/s3_log.h
```diff
@@ -66,16 +70,10 @@ inline const char* s3_log_get_req_id(const std::string& requestid) {
...
     if (loglevel >= s3log_level) {                               \
-      char* s3_log_msg__ = nullptr;                              \
-      int s3_log_len__ =                                         \
-          asprintf(&s3_log_msg__, "[Req:%s] " fmt "\n",          \
-                   s3_log_get_req_id(requestid), ##__VA_ARGS__); \
-      if (s3_log_len__ > 0) {                                    \
-        if (s3_log_msg__[s3_log_len__ - 2] == '\n')              \
-          s3_log_msg__[s3_log_len__ - 1] = '\0';                 \
-        s3_log_msg_##loglevel(s3_log_msg__);                     \
-        free(s3_log_msg__);                                      \
```
knrajnambiar76, line 77: no need of calling free(), that great

server/s3_log.cc
```diff
@@ -21,6 +21,18 @@
...
+enum {
+  KB = 1 << 10,
+};
+
+static thread_local char log_buffer[10*KB];
+char *__log_buff() {
+  return log_buffer;
+}
+size_t __log_buff_sz() {
+  return 10*KB - 1;
```
johnbent, line 33: If we are avoiding sizeof() for performance reasons, that's fine but the 10\*KB needs to be in a shared variable.  Having it hard-coded twice in the code is just asking for a future bug if someone changes it in one location and forgets to change it in the other.  I realize there are right next to each other so it might not happen but you could imagine that someone one day adds some other code between these two locations and that error becomes increasingly likely.

server/s3_log.h
johnbent: What happens if it does call free(), will it seg fault?

server/s3_log.cc
knrajnambiar76: defining a macro for 10*KB will be a good option, that macro can be used at both places

server/s3_log.h
knrajnambiar76: asnprintf needs free of s3_log_msg__ as that will be doing internal allocation from heap
with snprintf() its not needed, static array is passed to it.
Mostly compiler wont allow free of non heap allocated memory

server/s3_log.cc
johnbent: Or the enum that defines KB could just do this:

enum {

TEN\_KB = 10 \* (1 \<\< 10),

};

Also, is 1 \<\< 10 unnecessarily complex?  Why not just 

TEN\_KB = 10 \* 1024

server/s3_iem.h
```diff
@@ -34,22 +34,28 @@ extern S3Option* g_option_instance;
...
-           timestamp.c_str(), g_option_instance->get_s3_nodename().c_str(), \
-           getpid(), __FILE__, __LINE__, ##__VA_ARGS__);                    \
-    if (loglevel == LOG_ALERT) {                                            \
-      s3_syslog(loglevel, "IEC:AS" event_code ":" event_description);       \
-    } else {                                                                \
-      s3_syslog(LOG_ERR, "IEC:ES" event_code ":" event_description);        \
-    }                                                                       \
+#define s3_iem_(loglevel, s3_loglevel, event_code, event_desc, json_fmt, ...) \
+  do {                                                                        \
+    if (loglevel == LOG_ALERT) {                                              \
```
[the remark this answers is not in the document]
just-now: Sorry for late response.
This makes semantics of s3_iem() similar to s3_log(). In previous version, in case of S3_FATAL code would stop without writing IEMs into log.

server/s3_iem.h
```diff
@@ -34,22 +34,28 @@ extern S3Option* g_option_instance;
...
+      s3_syslog(loglevel, "IEC:AS" event_code ":" event_desc);                \
+    } else {                                                                  \
+      s3_syslog(LOG_ERR, "IEC:ES" event_code ":" event_desc);                 \
+    }                                                                         \
+    std::string timestamp = s3_get_timestamp();                               \
+    s3_log(S3_ ## s3_loglevel, "",                                            \
```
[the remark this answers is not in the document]
just-now: Can be avoided by means of const-expressions, I don't like macro deduction techniques either, still... sometimes CAT+CAT2 is a method to deduce and wrap things. Example: http://coliru.stacked-crooked.com/a/d8a896ad3563d403

server/s3_log.cc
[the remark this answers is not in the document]
just-now: This code is a dirty prototype. Don't think about it as a full fledged implementation. For sure it can be related by sizeof() and it'll not be slower than 10*KB-1 as it's a compile time constant.

server/s3_iem.h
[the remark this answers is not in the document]
just-now: BTW, in new version, `constexpressions` can be used instead.

server/s3_log.cc
just-now: `10*KB - 1` a compile-time constant and it will not hit performance.

server/s3_log.h
just-now: free() has constant complexity, allocation can be heavy in cases of memory pressure.

PR #832, opened 2021-04-14
EOS-18207: bucket metadata cache - part 2

evgeniy-brazhnikov:

### DIFF
--- a/server/s3_bucket_metadata_cache.cc
+++ b/server/s3_bucket_metadata_cache.cc
@@ -29,7 +29,7 @@
 #include "s3_log.h"
 #include "s3_request_object.h"
 
-S3BucketMetadataCache* p_bucket_metadata_cache;
+S3BucketMetadataCache* S3BucketMetadataCache::p_instance;
 
 std::unique_ptr<S3BucketMetadataV1>
 S3MotrBucketMetadataFactory::create_motr_bucket_metadata_obj(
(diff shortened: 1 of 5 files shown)


server/s3_bucket_metadata_cache.cc
```diff
@@ -45,53 +45,61 @@ S3BucketMetadataCache::Item::Item(const S3BucketMetadata& src)
 }
 
 S3BucketMetadataCache::Item::~Item() {
-  assert(can_remove());
-  assert(fetch_waiters.empty());
-  assert(!on_changed);
+  if (!can_remove()) {
+    s3_log(S3_LOG_FATAL, nullptr, "Motr operation is in progress");
```
knrajnambiar76, line 49: which scenario this case can arise ?

evgeniy-brazhnikov: It's for detecting possible logic errors in future

server/s3_bucket_metadata_cache.h
```diff
@@ -51,24 +54,29 @@ class S3BucketMetadataCache {
 
   virtual ~S3BucketMetadataCache();
 
+  static S3BucketMetadataCache* get_instance();
+
   void disable() { disabled = true; }
   void enable() { disabled = false; }
 
   using FetchHanderType =
```
knrajnambiar76, line 62: have comments on it and some other places also

knrajnambiar76 (approved):

server/s3_bucket_metadata_cache.h
evgeniy-brazhnikov: Will add.

server/s3_bucket_metadata_cache.cc
amitwac2608: we need to give more info in the log message, to figure out what went wrong... 

_"Motr operation is in progress" is not enough._

Give info like _"during X, Y happened, which is not expected. Please check or fix Z."_

knrajnambiar76: More message can be given, which is good

better not to say to "fix Z" unless we know exactly how that can be fixed.

amitwac2608 (approved):

PR #835, opened 2021-04-15
EOS-17303: S3faulttolerence - Object Del (single object) API support

DPG17:
Note: Existing UT and ST are successful.

### DIFF
--- a/s3config.release.yaml.sample
+++ b/s3config.release.yaml.sample
@@ -73,7 +73,7 @@ S3_SERVER_CONFIG:                                       # Section for S3 Server
    S3_BUCKET_METADATA_CACHE_MAX_SIZE: 10000             # Max count of entries in bucket MD cache
    S3_BUCKET_METADATA_CACHE_EXPIRE_SEC: 5               # Expiration time for bucket metadata in cache
    S3_BUCKET_METADATA_CACHE_REFRESH_SEC: 4              # Refresh timeout. After this timeout proactive MD re-load will happen.
-   S3_MAX_EXTENDED_OBJECTS_IN_FAULT_MODE: 0		# Maximum number of allowed recovery by S3 when fault happens and Motr error out with specific error code. When 0, feature is disabled. The maximum value is 10.
+   S3_MAX_EXTENDED_OBJECTS_IN_FAULT_MODE: 5		        # Maximum number of allowed recovery by S3 when fault happens and Motr error out with specific error code. When 0, feature is disabled. The maximum value is 5.
 S3_AUTH_CONFIG:                                         # Section for S3 Auth Service
    S3_AUTH_IP_ADDR: ipv4:127.0.0.1                      # Auth server IP address Should be in below format:
                                                         # ipv4 address format: ipv4:127.0.0.1
--- a/server/s3_delete_object_action.cc
+++ b/server/s3_delete_object_action.cc
@@ -121,7 +121,7 @@ void S3DeleteObjectAction::fetch_object_info_failed() {
 
 void S3DeleteObjectAction::add_object_oid_to_probable_dead_oid_list() {
   s3_log(S3_LOG_INFO, stripped_request_id, "%s Entry\n", __func__);
-
+  std::map<std::string, std::string> probable_oid_list;
   oid_str = S3M0Uint128Helper::to_string(object_metadata->get_oid());
   if (!motr_kv_writer) {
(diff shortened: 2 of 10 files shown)


s3config.release.yaml.sample
```diff
@@ -70,6 +70,7 @@ S3_SERVER_CONFIG:                                       # Section for S3 Server
    S3_SERVER_MOTR_ETIMEDOUT_MAX_THRESHOLD: 5            # Number of ETIMEDOUT errors per monitoring window before s3server restart
    S3_SERVER_MOTR_ETIMEDOUT_WINDOW_SEC: 60              # Monitoring window for motr ETIMEDOUT errors in seconds
    S3_SERVER_ENABLE_ADDB_DUMP: true                     # If set to true, then addb dump will be collected
+   S3_MAX_EXTENDED_OBJECTS_IN_FAULT_MODE: 0		# Maximum number of allowed recovery by S3 when fault happens and Motr error out with specific error code. When 0, feature is disabled. The maximum value is 10.
```
knrajnambiar76, line 73: maximum we can reduce to 5. 10 is high

server/s3_delete_object_action.cc
```diff
@@ -273,15 +333,42 @@ void S3DeleteObjectAction::delete_object() {
...
-    motr_writer = motr_writer_factory->create_motr_writer(
-        request, object_metadata->get_oid());
+  unsigned int object_count = del_object_oids.size();
+  if (object_count > 0) {
+    // process to delete object
+    std::pair<struct m0_uint128, int>& obj_layout_pair = del_object_oids.back();
+    if (!motr_writer) {
+      motr_writer = motr_writer_factory->create_motr_writer(
+          request, obj_layout_pair.first);
+    } else {
```
knrajnambiar76, line 343: is this else needed ?

s3config.release.yaml.sample
DPG17: This is configurable. we can always set to any value. I can set it to 5 on your request, but I don't know why only 5, why not 8 or 10 itself.

server/s3_delete_object_action.cc
DPG17: yes, else part is required to set correct oid to be deleted. Now, we have responsibility to delete multiple objects.

s3config.release.yaml.sample
knrajnambiar76: In calls we said that we will be supporting one or twice failure, 10 is pretty big. Have it at lower value

server/s3_delete_object_action.cc
knrajnambiar76: But if motr_writer is not set then create_motr_writer() instantiate it after that setting of OID needed right

knrajnambiar76 (approved):

s3config.release.yaml.sample
DPG17: By default, value is 0 to make feature disabled. But, I've added in the comment of this property to set max to 5.

server/s3_delete_object_action.cc
```diff
@@ -130,8 +130,12 @@ void S3DeleteObjectAction::add_object_oid_to_probable_dead_oid_list() {
 
   // prepending a char depending on the size of the object (size based bucketing
   // of object)
-  S3CommonUtilities::size_based_bucketing_of_objects(
-      oid_str, object_metadata->get_content_length());
+  size_t object_size = object_metadata->get_content_length();
+  if (object_metadata->is_object_extended()) {
```
amitwac2608, line 134: should we add else block and read the content length there, instead of doing that in above line ?

amitwac2608 (commented):
looks good.

server/s3_delete_object_action.cc
DPG17: well, no need of else block. The logic is first get size of object using get_content_length(). This should be sufficient for 95% of time, and we don't have to execute code within 'if' block. Only when object is extended, which is rare, we have to read object size using 'get_primary_obj_size()'. Moreover, the code also looks readable.

amitwac2608 (approved):

server/s3_delete_object_action.cc
DPG17: Actually, there is no much difference in the way you suggested here. So, any way can be done. My suggestion is let it be as is, .

nileshgovande:
Please run jenkins job

PR #836, opened 2021-04-15
[DONOT DELETE][DO NOT MERGE] EOS-19189 : Create configure_ldap API

shalakadharap:


scripts/ldap/base_configure_ldap.py
```diff
@@ -0,0 +1,155 @@
...
+if totalArguments > 5 and sys.argv[5] == forceclean :
+    forceclean = True
+
+INSTALLDIR="/opt/seagate/cortx/s3/install/ldap"
+#cleanup
+# Removing schemas
+try:
+    os.remove('/etc/openldap/slapd.d/cn=config/cn=schema/cn={1}s3user.ldif')
+except:
+    print('Error while deleting s3user.ldif file')
```
bkirunge7, line 38: In all print messages please print filename with full path

scripts/ldap/base_configure_ldap.py
```diff
@@ -0,0 +1,155 @@
...
+try:
+    os.remove('/etc/openldap/slapd.d/cn=config/olcDatabase={2}mdb.ldif')
+except:
+    print('Error while deleting mdb file')
+#Data Cleanup
+try:
+   if forceclean == True :
+        os.rmdir('/var/lib/ldap/')
+except:
+   print('Error while deleting ldap directory')
```
bkirunge7, line 66: print directory name with full path

scripts/ldap/base_configure_ldap.py
```diff
@@ -0,0 +1,155 @@
...
+l.simple_bind_s("cn=admin,dc=seagate,dc=com",ROOTDNPASSWORD)
+
+dn="dc=seagate,dc=com"
+add_record = [
+ ('dc', [b'seagate'] ),
+ ('o', [b'seagate'] ),
+ ('description', [b'Root entry for seagate.com.']),
+ ('objectClass', [b'top',b'dcObject',b'organization'])
+]
+l.add_s(dn, add_record)
```
bkirunge7, line 152: can you catch exception of these ldap api calls and print message with detail about step that failed

scripts/ldap/base_configure_ldap.py
shalakadharap: Fixed

scripts/ldap/base_configure_ldap.py
shalakadharap: Fixed

scripts/ldap/base_configure_ldap.py
shalakadharap: Fixed

sachitanands (approved):

scripts/ldap/base_configure_ldap.py
```diff
@@ -0,0 +1,217 @@
...
+defaultpasswds="ldapadmin"
+
+if totalArguments <= 1 :
+    print('Too few arguments.' , usage)
+    sys.exit()
+if totalArguments > 1 :
+    LDAPADMINPASS =  sys.argv[1]
+if totalArguments > 2 :
+    ROOTDNPASSWORD = sys.argv[2]
+if totalArguments > 3 and sys.argv[3] == usedefaultpassword :
```
abhilekh268, line 44: is it the case here usedefaultpassword should be given as third argument only?

bkirunge7 (approved):

scripts/ldap/base_configure_ldap.py
shalakadharap: Yes.Order needs to be preserved

abhilekh268: can we use argument parser here something like -p password like this.....order need not be preserved and one more query sys.arg[3] we are checking for userpassword it has to be always at 3rd position?

shalakadharap: YEs.Order needs to be preserved

shalakadharap: I need to check on argument parser as I am not much aware of python.

abhilekh268: better if we use argument parser,this will be hard coded....otherwise and userpassword has to be always 3rd argument what if i don;t pass first two...and pass userdefaultpassword as first it won;t process.we have used argument parser in s3iamcli i guess u can take reference

shalakadharap: @abhilekh268 Implemented argument parser. Working fine in any order now. Check 1s.

abhilekh268 (approved):

scripts/ldap/base_configure_ldap.py
```diff
@@ -0,0 +1,209 @@
...
+
+mod_attrs = [( ldap.MOD_ADD, 'olcAccess', [b'{1}to * by dn.base="cn=sgiamadmin,dc=seagate,dc=com" write by self write by * none'] )]
+try:
+    l.modify_s(dn,mod_attrs)
+except:
+    print('Error while modifying olcAccess attribute for * and dn.base=cn=sgiamadmin,dc=seagate,dc=com in olcDatabase={2}mdb')
+
+l.unbind_s()
+
+os.system('systemctl restart slapd')
```
AjinkyaDhumal, line 187: I think @SaumitraK531 recently removed some slapd restarts. Now we have only 1 restart after initial cleanup. Can you check if that can be addressed here.

AjinkyaDhumal (changes requested):

scripts/ldap/base_configure_ldap.py
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env python3
```
AjinkyaDhumal, line 1: This file needs to be merged in cortx-utils repo. Lets raise a separate PR on cortx-utils repo, Sachin Punadikar will be merging it to utils. We can add approvals there.

AjinkyaDhumal (changes requested):

scripts/ldap/base_configure_ldap.py
shalakadharap: Sure. Will check and see if it can be removed

scripts/ldap/base_configure_ldap.py
shalakadharap: ok

scripts/ldap/base_configure_ldap.py
shalakadharap: fixed

scripts/ldap/base_configure_ldap.py
shalakadharap: https://github.com/Seagate/cortx-utils/pull/262
Raised.
Pls check and if ok then add approvals there

gowthamchinna:
test this please

PR #848, opened 2021-04-21
EOS-19156: 's3_setup config' fails in 3 node miniprovisioner setup

SaumitraK531:
Signed-off-by: Saumitra Kulkarni <saumitra.kulkarni@seagate.com>

### DIFF
--- a/scripts/ldap/setup_ldap.sh
+++ b/scripts/ldap/setup_ldap.sh
@@ -80,14 +80,15 @@ done
 INSTALLDIR="/opt/seagate/cortx/s3/install/ldap"
 yum list installed selinux-policy && yum update -y selinux-policy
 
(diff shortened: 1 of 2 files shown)


scripts/ldap/setup_ldap.sh
```diff
@@ -87,6 +87,7 @@ rm -f /etc/openldap/slapd.d/cn\=config/cn\=schema/*ppolicy.ldif
 # Removing configuration files
```
sachitanands, line 87: Please add a comment mentioning that we need to delete all prev loaded modules else we may see opendap idempotency issue

sachitanands (approved):

amitwac2608 (approved):

PR #853, opened 2021-04-22
EOS-19564:Auth:Fix for Authserver LogLevel values

sachitanands:
kesytore.properties does not required logger config which was causing conflicts for "logLevel" properteis with authserver.properties 
hence debug mode value was updated as INFO while loading authserver 

### DIFF
--- a/auth/resources/keystore.properties.sample
+++ b/auth/resources/keystore.properties.sample
@@ -24,14 +24,3 @@ s3KeyStorePassword=seagate
 s3KeyPassword=seagate
 s3AuthCertAlias=s3auth_pass
 s3CipherUtil=s3cipher generate_key --const_key cortx
-
-# Uncomment this to override logging config file.
-#
-# logConfigFile=<PATH TO CONFIG FILE>
-
-# Uncomment this to override log level.
-# Note - setting this will override the logging level of root logger.
-# If this option is set along with logConfigFile, the log level set in the
-# config file will be overridden.
-#
-logLevel=INFO


shalakadharap (approved):
Just make sure its not breaking SW Update (Check with deepak 1s)

sachitanands:
> Not sure why its added here but if its bcz of SW update then make sure its not breaking (Check with deepak 1s)

yes. it was already there from beginning in commented mode but with recent changes of confstore and s/w update it was uncommented and it created this problem.

abhilekh268 (approved):

auth/resources/keystore.properties.sample
```diff
@@ -24,14 +24,3 @@ s3KeyStorePassword=seagate
...
-# Uncomment this to override logging config file.
-#
-# logConfigFile=<PATH TO CONFIG FILE>
-
-# Uncomment this to override log level.
-# Note - setting this will override the logging level of root logger.
-# If this option is set along with logConfigFile, the log level set in the
-# config file will be overridden.
-#
-logLevel=INFO
```
abhilekh268, line 37: What will be default loglevel?if we don;t have this entry itself?and we don;t need to manipulate loglevel?

sachitanands: the default log level will be "INFO" only and it can be changed from authserver.properties and there is no change in this .
We dont need entry in keystore.properties ..hence i have removed this form keystore.properties

nileshgovande (approved):

PR #855, opened 2021-04-22
 EOS-19192: Remove control character dependency and add extended metadata to different index table

DPG17:


DPG17:
Was opened in an error. So, closing

PR #866, opened 2021-04-27
github833: S3 logger robustness, performance improvement and reorg proposal

vimalraghwani20:
- removed ASPRINTF! and unneeded allocations for EVERY! log write!!!
- uses tls-approach.
- change the semantics of s3_iem() so that it becomes similar to s3 log.
- use templated verbosity select method to be able to handle const-expressions inside s3_log macro
- example: s3_log(cond ? S3_LOG_DEBUG : S3_LOG_ERROR, ...).

### DIFF
--- a/server/s3_log.cc
+++ b/server/s3_log.cc
@@ -21,11 +21,9 @@
 #include "s3_log.h"
 #include "s3_option.h"
 
-enum {
-  KB = 1 << 10,
-};
+#define ONE_KB (1 << 10)
 
-static thread_local char log_buffer[10 * KB];
+static thread_local char log_buffer[10 * ONE_KB];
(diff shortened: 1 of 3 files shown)


server/s3_log.cc
```diff
@@ -21,6 +21,14 @@
 #include "s3_log.h"
 #include "s3_option.h"
 
+enum {
+  KB = 1 << 10,
```
knrajnambiar76, line 25: you can ONEKB as macro

vimalraghwani20: fixed

knrajnambiar76 (approved):

nileshgovande (approved):

PR #868, opened 2021-04-28
EOS-19814:Auth: increase LdapResult size from 500 -> 10000 for dev vm

sachitanands:
In case of dev vm we have max ldap search limit set as default value ie.e. 500 
hence if we have entries more than 500 then create/list API will be broken with "Size Limit Exceeded" error thrown by ldap.

In case of provisioned setup:
This case is already handled in setup_ldap.sh script hence no changes required in case of prod setup script.

### DIFF
--- a/ansible/openldap.yml
+++ b/ansible/openldap.yml
@@ -104,6 +104,11 @@
           src: files/ldap/s3slapdindex.ldif
           dest: /tmp/s3ldap/s3slapdindex.ldif
 
+      - name: Copy s3 ldap resultssizelimit file
+        template:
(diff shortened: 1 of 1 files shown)


sachitanands:
Tested ansible work by running pre-merge jobs.
image

sachitanands:
mini-provisioner job has failed because of infra issue.
retriggered new job : http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Mini-Provisioning/job/cortx-s3server/925 (In-Progress)

shalakadharap (approved):

nileshgovande (approved):

PR #878, opened 2021-05-03
EOS-20265:add missing path in PATH variable

sachitanands:
In case of dev vm we are observing the environment variable PATH which does not has "/usr/local/bin" entry hence python modules in s3server like s3iamcli, s3confstore, s3cipher seems to be broken.

### DIFF
--- a/jenkins-build.sh
+++ b/jenkins-build.sh
@@ -381,6 +381,10 @@ if [ $cleanup_only -eq 1 ]; then
   exit
 fi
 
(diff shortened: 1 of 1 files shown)


vimalraghwani20 (approved):

sachitanands:
mini-prov job :  http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Mini-Provisioning/job/cortx-s3server/948

PR #882, opened 2021-05-04
EOS-17955: Multipart abort

knrajnambiar76:


server/s3_abort_multipart_action.cc
```diff
@@ -382,39 +484,76 @@ void S3AbortMultipartAction::startcleanup() {
...
+void S3AbortMultipartAction::delete_parts_successful() {
   s3_log(S3_LOG_INFO, stripped_request_id, "%s Entry\n", __func__);
+  if (multipart_parts.size() == 0) {
+    return;
+  }
+  s3_log(S3_LOG_INFO, request_id,
+         "Deleted old object oid "
+         "[%" SCNx64 " : %" SCNx64 "]",
+         (motr_writer->get_oid()).u_hi, (motr_writer->get_oid()).u_lo);
+  multipart_parts.erase(multipart_parts.begin());
```
DPG17, line 544: Erasing from beginning of vector would lead to acquiring new memory and copying elements. This would be significant, impact if number of elements in it is quite high (say, 10 thousand parts, which is AWS limit on number of parts)

DPG17 (approved):

server/s3_abort_multipart_action.cc
knrajnambiar76: Make sense, will change that

knrajnambiar76: extracting from back and deleting from end

amitwac2608 (approved):

PR #894, opened 2021-05-11
EOS-20539: Fix argparse known issue to except dash('-') in option value.

amitwac2608:
link to known issue: https://bugs.python.org/issue9334

### DIFF
--- a/scripts/s3-support-bundles/s3_bundle_generate.sh
+++ b/scripts/s3-support-bundles/s3_bundle_generate.sh
@@ -43,8 +43,8 @@ sgiamadminpwd=''
 constkey=cortx
 propertiesfilepath="properties:///opt/seagate/cortx/auth/resources/authserver.properties"
 ldapkey="ldapLoginPW"
-ldapcipherkey=$(s3cipher generate_key --const_key "$constkey")
-encryptedkey=$(s3confstore $propertiesfilepath getkey --key="$ldapkey")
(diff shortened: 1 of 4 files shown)


scripts/s3-support-bundles/s3_bundle_generate.sh
```diff
@@ -43,11 +43,11 @@ sgiamadminpwd=''
 constkey=cortx
 propertiesfilepath="properties:///opt/seagate/cortx/auth/resources/authserver.properties"
 ldapkey="ldapLoginPW"
-ldapcipherkey=$(s3cipher generate_key --const_key $constkey)
-encryptedkey=$(s3confstore $propertiesfilepath getkey --key $ldapkey)
+ldapcipherkey=$(s3cipher generate_key --const_key "$constkey")
+encryptedkey=$(s3confstore $propertiesfilepath getkey --key="$ldapkey")
```
nileshgovande, line 47: This is OK as a quick-fix - can we make this change everywhere in s3cipher and  s3confstore calls (and in help as well) and help to be bullet-proof?

sachitanands (approved):

scripts/s3-support-bundles/s3_bundle_generate.sh
amitwac2608: I did grep in the repo, and included all files in this PR which use s3confstore and s3cipher, except "install_prereqs.sh" file. Including in next commit.

PR #922, opened 2021-05-24
[EOS-20765] Added Size check in multipart upload

nileshgovande:


server/s3_put_multiobject_action.cc
```diff
@@ -134,6 +134,12 @@ void S3PutMultiObjectAction::check_part_details() {
   } else if ((request->get_object_name()).length() > MAX_OBJECT_KEY_LENGTH) {
     set_s3_error("KeyTooLongError");
     send_response_to_s3_client();
+  } else if (request->get_data_length() <= MINIMUM_ALLOWED_PART_SIZE) {
```
knrajnambiar76, line 137: This check for MINIMUM_ALLOWED_PART_SIZE can  be a problem if the part happens to be the last one
last part can be of size less than MINIMUM_ALLOWED_PART_SIZE

server/s3_put_multiobject_action.cc
```diff
@@ -134,6 +134,12 @@ void S3PutMultiObjectAction::check_part_details() {
...
+    set_s3_error("EntityTooSmall");
+    send_response_to_s3_client();
+  } else if (request->get_data_length() >= MAXIMUM_ALLOWED_PART_SIZE) {
```
knrajnambiar76, line 140: get_content_length() will get the content length

nileshgovande:
All 6 checks passed - Merging

PR #931, opened 2021-05-25
EOS-20605: CopyObject regression

DPG17:
- Code changes to trim buffer after the source object size(during copy operation)

Note: MD5 checksum of downloaded object is still not matching with the original source object, and investigation for the same is in progress.

### DIFF
--- a/server/s3_object_data_copier.cc
+++ b/server/s3_object_data_copier.cc
@@ -39,6 +39,7 @@ S3ObjectDataCopier::S3ObjectDataCopier(
 
   request_id = this->request_object->get_request_id();
   size_of_ev_buffer = g_option_instance->get_libevent_pool_buffer_size();
+  assert(read_data_buffer.size() == 0);
   s3_log(S3_LOG_DEBUG, request_id, "%s Ctor\n", __func__);
 }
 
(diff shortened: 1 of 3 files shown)


server/s3_object_data_copier.cc
```diff
@@ -122,38 +131,23 @@ void S3ObjectDataCopier::read_data_block_success() {
...
-    return;
-  }
-  bytes_left_to_read -= bytes_in_chunk_count;
-
-  auto& r_last_block_size =
-      data_blocks_read[data_blocks_read.size() - 1].second;
-
-  if (r_last_block_size >= bytes_left_to_read) {
-    r_last_block_size = bytes_left_to_read;
+    get_buffers(data_blocks_read, bytes_in_chunk_count, bytes_left_to_read);
```
DPG17, line 145: This function call (get_buffers()) is needed, otherwise the ETAG result of CopyObject api does not match the md5sum stored for the object in object list index table

knrajnambiar76, line 145: Padding in writer will pad with same buffer, so releasing here will save space

knrajnambiar76 (approved):

evgeniy-brazhnikov (approved):

server/s3_object_data_copier.cc
DPG17: ok. gr8

PR #939, opened 2021-05-27
Anonymous user is able to perform Copy object operation

shalakadharap:


shalakadharap:
Unwanted and Wrongly raised

PR #957, opened 2021-06-04
EOS20850-User should not get Logged out after trying to delete an inv…

shalakadharap:
…alid user


bkirunge7 (commented):
Add a system test to cover this

shalakadharap:
> Add a system test to cover this

We wont be able to handle this in STs because we are not changing error output here
we are just changing http status code so response will look as it was before-
image

sachitanands (approved):

bkirunge7 (approved):

vimalraghwani20:
All 5 checked have passed.

PR #976, opened 2021-06-11
EOS-17572: S3 Multipart Post Complete api implementation (Multipart re-design)

DPG17:
This change also has other fixes e.g.:
1. Get Object api changes to read several objects created as part of multipart redesign.
    These code change are ported from FaultTolerance branch, with some changes.
2. Changes in implementation files related to Multipart functionality, other than Complete multipart API file

### DIFF
--- a/server/s3_get_object_action.h
+++ b/server/s3_get_object_action.h
@@ -32,6 +32,27 @@
 #include "s3_factory.h"
 #include "s3_timer.h"
 
+struct S3ExtendedObjectInfo {
+ public:
+  S3ExtendedObjectInfo() {
+    start_offset_in_object = 0;
+    object_OID = {0, 0};
+    object_layout = -1;
+    object_pvid = {0, 0};
+    total_blocks_in_object = 0;
+    object_size = 0;
+    requested_object_size = 0;
+  }
+  size_t start_offset_in_object;
+  size_t total_blocks_in_object;
+  size_t object_size;
+  // Actual length to be returned to client
+  size_t requested_object_size;
+  struct m0_uint128 object_OID;
+  int object_layout;
+  struct m0_fid object_pvid;
+};
+
 class S3GetObjectAction : public S3ObjectAction {
 
   std::shared_ptr<S3MotrReader> motr_reader;
@@ -49,6 +70,14 @@ class S3GetObjectAction : public S3ObjectAction {
   bool read_object_reply_started;
   std::shared_ptr<S3MotrReaderFactory> motr_reader_factory;
   S3Timer s3_timer;
+
+  // TODO: Data structures for extended object type
+  size_t total_objects;
+  size_t total_objects_to_read;
+  unsigned int next_fragment_object;
+  std::vector<struct S3ExtendedObjectInfo> extended_objects;
+  size_t total_blocks_to_read_all_objects;
+  size_t data_sent_to_client_for_object;
 
   size_t get_requested_content_length() const {
     return last_byte_offset_to_read - first_byte_offset_to_read + 1;
@@ -69,9 +98,12 @@ class S3GetObjectAction : public S3ObjectAction {
   void validate_object_info();
   void check_full_or_range_object_read();
   void set_total_blocks_to_read_from_object();
+  void set_total_blocks_to_read_from_fragmented_object();
+
   bool validate_range_header_and_set_read_options(
       const std::string& range_value);
   void read_fragmented_object();
+  void set_total_blocks_to_read_from_next_object();
   void read_object();
 
   void read_object_data();
--- a/server/s3_object_action_base.cc
+++ b/server/s3_object_action_base.cc
@@ -106,33 +106,43 @@ void S3ObjectAction::fetch_object_info() {
 }
 
 void S3ObjectAction::fetch_object_info_success() {
+  s3_log(S3_LOG_DEBUG, request_id, "%s Entry\n", __func__);
   request->set_object_size(object_metadata->get_content_length());
   // TODO: Read extended object's parts/fragments, depending on the type of
   // primary object.
   // If object is extended, create S3ObjectExtendedMetadata and load extended
   // entries.
   if (object_metadata->is_object_extended()) {
-    // Read the extended parts of the object
-    extended_obj_metadata =
+    // Read the extended parts of the object from extended index table
+    std::shared_ptr<S3ObjectExtendedMetadata> extended_obj_metadata =
         object_metadata_factory->create_object_ext_metadata_obj(
             request, request->get_bucket_name(), request->get_object_name(),
             object_metadata->get_obj_version_key(),
             object_metadata->get_number_of_parts(),
-            object_metadata->get_number_of_fragments(), object_list_oid);
-
+            object_metadata->get_number_of_fragments(),
+            bucket_metadata->get_extended_metadata_index_oid());
+    object_metadata->set_extended_object_metadata(extended_obj_metadata);
     extended_obj_metadata->load(
         std::bind(&S3ObjectAction::fetch_ext_object_info_success, this),
         std::bind(&S3ObjectAction::fetch_ext_object_info_failed, this));
   } else {
     next();
   }
+  s3_log(S3_LOG_DEBUG, request_id, "%s Exit\n", __func__);
 }
 
-void S3ObjectAction::fetch_ext_object_info_success() { next(); }
+void S3ObjectAction::fetch_ext_object_info_success() {
+  s3_log(S3_LOG_DEBUG, request_id, "%s Entry\n", __func__);
+  next();
+  s3_log(S3_LOG_DEBUG, request_id, "%s Exit\n", __func__);
+}
 
 void S3ObjectAction::fetch_ext_object_info_failed() {
-  // TBD: Add code to handle failure in loading extended entries.
+  // Add code in derived class to better handle
+  // failure in loading extended entries.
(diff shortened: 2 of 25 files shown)


server/s3_object_action_base.cc
```diff
@@ -106,33 +106,42 @@ void S3ObjectAction::fetch_object_info() {
...
-void S3ObjectAction::fetch_ext_object_info_success() { next(); }
+void S3ObjectAction::fetch_ext_object_info_success() {
+  s3_log(S3_LOG_DEBUG, request_id, "%s Entry\n", __func__);
+  next();
+  s3_log(S3_LOG_DEBUG, request_id, "%s Exit\n", __func__);
+}
 
 void S3ObjectAction::fetch_ext_object_info_failed() {
   // TBD: Add code to handle failure in loading extended entries.
+  s3_log(S3_LOG_DEBUG, request_id, "%s Entry\n", __func__);
```
knrajnambiar76, line 142: Failure scenario will be handled separately in which ticket ?

knrajnambiar76 (approved):

server/s3_object_action_base.cc
DPG17: need to implement in derived class...will do this soon.

server/s3_get_object_action.h
```diff
@@ -32,6 +32,27 @@
 #include "s3_factory.h"
 #include "s3_timer.h"
 
+struct S3ExtendedObjectInfo {
+ public:
+  S3ExtendedObjectInfo() {
+    start_offset_in_object = 0;
+    object_OID = {0, 0};
+    object_layout = -1;
+    object_pvid = {0, 0};
```
evgeniy-brazhnikov, line 41: It can slash with my changes which are already merged to main branch.
Object pvid is already handled

server/s3_object_action_base.cc
DPG17: done. Pushed changes

server/s3_object_metadata.cc
```diff
@@ -822,11 +809,15 @@ int S3ObjectMetadata::from_json(std::string content) {
     primary_obj_size = newroot["Size"].asUInt64();
   }
 
-  if (obj_fragments == 0 && obj_parts != 0) {
+  if (obj_fragments != 0 && obj_parts != 0 && (obj_fragments == obj_parts)) {
```
evgeniy-brazhnikov, line 812: (obj_fragments == 0 ...) in old code.
Is this correct?

server/s3_post_complete_action.cc
```diff
@@ -155,6 +149,29 @@ void S3PostCompleteAction::fetch_object_info_failed() {
   s3_log(S3_LOG_DEBUG, stripped_request_id, "Exiting\n");
 }
 
+void S3PostCompleteAction::fetch_ext_object_info_failed() {
+  s3_log(S3_LOG_DEBUG, request_id, "%s Entry\n", __func__);
+  if (object_metadata && object_metadata->is_object_extended()) {
```
evgeniy-brazhnikov, line 154: It seams that send_response_to_s3_client() must be moved outside the condition

server/s3_post_complete_action.cc
```diff
@@ -424,77 +440,152 @@ bool S3PostCompleteAction::validate_parts() {
...
+  if (new_object_ext_metadata) {
+    // Add object part to new object's extended metadata
+    struct s3_part_frag_context part_frag_ctx = {};
+    part_frag_ctx.motr_OID = part_metadata->get_oid();
+    part_frag_ctx.PVID = part_metadata->get_pvid();
+    part_frag_ctx.versionID = object_metadata->get_obj_version_key();
+    part_frag_ctx.item_size = part_metadata->get_content_length();
+    part_frag_ctx.layout_id = part_metadata->get_layout_id();
+    part_frag_ctx.is_multipart = true;
+    const int part_number =
```
evgeniy-brazhnikov, line 503: Is there narrow cast?
A compiler or a static analizer can give warning

server/s3_post_complete_action.cc
```diff
@@ -424,77 +440,152 @@ bool S3PostCompleteAction::validate_parts() {
...
+// the specified probable dead oid list
+void S3PostCompleteAction::add_part_object_to_probable_dead_oid_list(
+    const std::shared_ptr<S3ObjectMetadata>& object_metadata,
+    std::vector<std::unique_ptr<S3ProbableDeleteRecord>>&
+        parts_probable_del_rec_list,
+    bool is_old_object) {
   s3_log(S3_LOG_INFO, stripped_request_id, "%s Entry\n", __func__);
+  const std::shared_ptr<S3ObjectExtendedMetadata>& object_ext_metadata =
+      object_metadata->get_extended_object_metadata();
+  unsigned int total_parts = 0;
```
evgeniy-brazhnikov, line 531: m.b.
unsigned total_parts = object_ext_metadata ? object_ext_metadata->get_part_count() : 0
would be more clear?

server/s3_post_complete_action.cc
```diff
@@ -424,77 +440,152 @@ bool S3PostCompleteAction::validate_parts() {
...
+  if (object_ext_metadata &&
+      ((total_parts = object_ext_metadata->get_part_count()) > 0)) {
+    s3_log(S3_LOG_DEBUG, request_id, "Total parts available [%u]\n",
+           total_parts);
+    // object has some parts
+    const std::map<int, std::vector<struct s3_part_frag_context>>&
```
evgeniy-brazhnikov, line 537: I would use 
const auto& ...
to avoid possible implicit type cast.

server/s3_post_complete_action.cc
```diff
@@ -424,77 +440,152 @@ bool S3PostCompleteAction::validate_parts() {
...
+        ext_entries_mp = object_ext_metadata->get_raw_extended_entries();
+    struct m0_uint128 old_oid, part_list_index_oid;
+    bool is_multipart = false;
+    for (unsigned int key_indx = 1; key_indx <= total_parts; key_indx++) {
+      const std::vector<struct s3_part_frag_context>& part_entry =
+          ext_entries_mp.at(key_indx);
```
evgeniy-brazhnikov, line 543: std::map<...>::at(...) can throw exceptions.
Can we be sure that this won't happen?

server/s3_post_complete_action.cc
```diff
@@ -505,76 +596,56 @@ void S3PostCompleteAction::add_object_oid_to_probable_dead_oid_list() {
...
+    // Mark old object if any for probable deletion.
+    // store old object oid
+    if (old_object_oid.u_hi || old_object_oid.u_lo) {
+      assert(!old_oid_str.empty());
+      assert(!new_oid_str.empty());
+      add_part_object_to_probable_dead_oid_list(
+          object_metadata, old_parts_probable_del_rec_list, true);
+      std::map<std::string, std::string> kv_list;
+      for (auto& probable_rec : old_parts_probable_del_rec_list) {
+        kv_list[probable_rec->get_key()] = probable_rec->to_json();
```
evgeniy-brazhnikov, line 631: Can we be sure, that all keys would be unique?

server/s3_post_complete_action.cc
```diff
@@ -933,48 +1004,110 @@ void S3PostCompleteAction::delete_old_object() {
     next();
     return;
   }
-  motr_writer->delete_object(
-      std::bind(&S3PostCompleteAction::remove_old_object_version_metadata,
-                this),
-      std::bind(&S3PostCompleteAction::next, this), old_object_oid,
-      old_layout_id, multipart_metadata->get_pvid());
+  unsigned int object_count = old_obj_oids.size();
```
evgeniy-brazhnikov, line 1007: Narrow cast

server/s3_post_complete_action.cc
DPG17: It can happen when 'total_parts' (which is the value for field "PARTS" in object list index) is more than actual number of entries in the extended md index for the same object with version. This is less likely to occur, until the entries are removed externally from extended md index for the object/version OR, there is some error in saving parts to part list index table during multipart upload api, which is less likely to occur because we return service unavailable or internal error, when there is any error in writing part entry to part index in multipart upload.

So, less likely to happen, but can occur. Will try to guard it using exception.

server/s3_post_complete_action.cc
DPG17: I agree with you. But let it be like this now.  This makes code more readable.

server/s3_post_complete_action.cc
DPG17: yes, will adopt your suggestion.

server/s3_post_complete_action.cc
DPG17: done

server/s3_post_complete_action.cc
DPG17: yes. key here would be unique...because they are OIDs

server/s3_post_complete_action.cc
DPG17: ok. This condition is true for such failed callback, but yes will remove if condition to avoid "no response to client" situation

server/s3_post_complete_action.cc
DPG17: ok. will use correct type

DPG17: Compiler did not shout here...not sure why?

evgeniy-brazhnikov (approved):

PR #982, opened 2021-06-14
EOS-20767: Check for PutObjectACL permission in target bucket for CopyObject API.

amitwac2608:
If user passes "x-amx-acl" header with CopyObject API request, s3server should check for the
required permission: PutObjectACL on target bucket, before start copying.


server/s3_copy_object_action.cc
```diff
@@ -381,7 +381,9 @@ void S3CopyObjectAction::set_authorization_meta() {
   if (!additional_object_metadata->get_tags().empty()) {
     request->set_action_list("PutObjectTagging");
   }
-  // request->set_action_list("PutObjectAcl");
+  if (!request->get_header_value("x-amz-acl").empty()) {
```
knrajnambiar76, line 384: why so many file changes for "Check for PutObjectACL...."

PR #988, opened 2021-06-16
[EOS-19710] Refactoring for PUT DI

knrajnambiar76:


server/s3_motr_wrapper.cc
```diff
@@ -171,6 +188,135 @@ int ConcreteMotrAPI::motr_idx_op(struct m0_idx *idx, enum m0_idx_opcode opcode,
 
 void ConcreteMotrAPI::motr_idx_fini(struct m0_idx *idx) { m0_idx_fini(idx); }
 
+#if 0
```
nileshgovande, line 191: Let us not disable it until we integrate with motr

server/s3_motr_wrapper.cc
```diff
@@ -188,6 +334,16 @@ int ConcreteMotrAPI::motr_obj_op(struct m0_obj *obj, enum m0_obj_opcode opcode,
     (*op)->op_sm.sm_state = M0_OS_INITIALISED;
     return 0;
   }
+
+/* Backup / Populate the attr */
+#if 0
```
nileshgovande, line 339: Let us not disable it until we integrate with motr

PR #990, opened 2021-06-17
[EOS-21134] Fix order parts' etags concatenated - make it explicit in order of parts

dsurnin:

### DIFF
--- a/server/s3_post_complete_action.h
+++ b/server/s3_post_complete_action.h
@@ -76,7 +76,6 @@ class S3PostCompleteAction : public S3ObjectAction {
   size_t prev_fetched_parts_size;
   size_t validated_parts_count;
   std::string last_key;
-  S3AwsEtag awsetag;
 
   struct m0_uint128 old_object_oid;
   int old_layout_id;
@@ -89,6 +88,9 @@ class S3PostCompleteAction : public S3ObjectAction {
   // Probable delete record for new object OID in case of current req failure
   std::string new_oid_str;  // Key for new probable delete rec
   std::unique_ptr<S3ProbableDeleteRecord> new_probable_del_rec;
+
+  std::map<unsigned int, std::string> part_etags;
+  std::string generate_etag();
 
  public:
   S3PostCompleteAction(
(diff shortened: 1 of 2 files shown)


server/s3_post_complete_action.h
```diff
@@ -90,6 +89,9 @@ class S3PostCompleteAction : public S3ObjectAction {
   std::string new_oid_str;  // Key for new probable delete rec
   std::unique_ptr<S3ProbableDeleteRecord> new_probable_del_rec;
 
+  std::map<unsigned int, std::string> part_etags;
```
bkirunge7, line 92: using vector or string array will be efficient with part number as index, compared to ordered map

dsurnin: According to aws doc
'''
Parts upload

When uploading a part, in addition to the upload ID, you must specify a part number. You can choose any part number between 1 and 10,000. A part number uniquely identifies a part and its position in the object you are uploading. The part number that you choose doesn’t need to be in a consecutive sequence (for example, it can be 1, 5, and 14). If you upload a new part using the same part number as a previously uploaded part, the previously uploaded part is overwritten. 
'''
due to discontinuity we will have to preallocate memory for all 10000 elements even in case we have only 2 parts: 1 and 10000.
in case of vector/array usage we will have to add checks if checksum we actually added with the specified index.
also in case of big holes in parts' indexes vector/array will be less efficient than map, example indexes: 1, 1000, 10000, these indexes are valid parts but with vector/array you will need to traverse all 10000 elements to find 3 of them.

however I believe examples above are extreme cases and probably we will not have such cases.
so, @bkirunge7 what do you prefer?
I can leave it as is
I can change it to vector
I can make a research how much effort we can gain by using vector or map in different scenarios

bkirunge7: Good analysis, if there are holes and all parts are not guaranteed to be present between 1..n, Then map would be better option

bkirunge7 (approved):

dsurnin:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/S3server-cortx-1.0/55/

t7ko-seagate (approved):

nileshgovande (approved):

PR #1017, opened 2021-07-02
[EOS19172] : Refactoring of backgrounddelete account creation process

pranavPawar07:
**Problem Statement/Overview:** 
The current approach is we are expecting the caller to send parameters of bgdelete for account creation.
similar things are observed in multiple phases e.g. initcmd.py, resetcmd.py

**Solution:**
1. Create a common function in setupcmd.py which returns a dict of bg params which is used to create the bg delete account in the init and reset phase.

Signed-off-by: Pranav Diliprao Pawar <pranav.pawar@seagate.com>

### DIFF
--- a/s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
+++ b/s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
@@ -67,3 +67,12 @@ leakconfig:                                  # Section for object leak config
    leak_processing_delay_in_mins: 15                          # Time delay in mins after which probable delete entry is procesed for leak.
    version_processing_delay_in_mins: 5                        # Time delay in mins after which object version is processed
    cleanup_enabled: False                                     # Flag for cleanup during S3 instance non-active.
+
+background_delete_account:
+   account_name: "s3-background-delete-svc"
+   account_id: "67891"
+   canonical_id: "C67891"
+   mail: "s3-background-delete-svc@seagate.com"
+   s3_user_id: "450"
+   const_cipher_secret_str: "s3backgroundsecretkey"
+   const_cipher_access_str: "s3backgroundaccesskey"
--- a/s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
+++ b/s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
@@ -68,11 +68,12 @@ leakconfig:                                  # Section for object leak config
    version_processing_delay_in_mins: 5                        # Time delay in mins after which object version is processed
    cleanup_enabled: False                                     # Flag for cleanup during S3 instance non-active.
 
+#DO NOT CHANGE ANY KEY NAME
 background_delete_account:
-   account_name: "s3-background-delete-svc"
-   account_id: "67891"
-   canonical_id: "C67891"
-   mail: "s3-background-delete-svc@seagate.com"
-   s3_user_id: "450"
-   const_cipher_secret_str: "s3backgroundsecretkey"
-   const_cipher_access_str: "s3backgroundaccesskey"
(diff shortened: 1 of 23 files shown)


s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
```diff
@@ -67,3 +67,12 @@ leakconfig:                                  # Section for object leak config
    leak_processing_delay_in_mins: 15                          # Time delay in mins after which probable delete entry is procesed for leak.
    version_processing_delay_in_mins: 5                        # Time delay in mins after which object version is processed
    cleanup_enabled: False                                     # Flag for cleanup during S3 instance non-active.
+
+background_delete_account:
```
vimalraghwani20, line 71: add comment sAYING DO NOT CHANGE KEY NAME. as it will impact creation of bg delete account.

s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
```diff
@@ -67,3 +67,12 @@ leakconfig:                                  # Section for object leak config
...
+   account_name: "s3-background-delete-svc"
```
vimalraghwani20, line 72: add comment for each variables like other config keys

scripts/provisioning/setupcmd.py
```diff
@@ -617,3 +618,16 @@ def get_ldap_root_credentials(self):
 
     if enc_rootdn_passwd != None:
       self.rootdn_passwd = s3cipher_obj.decrypt(cipher_key, enc_rootdn_passwd)
+
+  def create_dict_for_BG_delete_account(self):
```
vimalraghwani20, line 622: rename this function to something like get_config_param_for_BG_delete_account()

scripts/provisioning/setupcmd.py
```diff
@@ -617,3 +618,16 @@ def get_ldap_root_credentials(self):
...
+    """returns a dictionary required for creating a BG delete account in init and reset phase."""
+    opfileconfstore = S3CortxConfStore(f'yaml://{self.BG_delete_config_file}', 'read_config_idx')
```
vimalraghwani20, line 624: change index name to read_bg_delete_config_idx

scripts/provisioning/setupcmd.py
```diff
@@ -617,3 +618,16 @@ def get_ldap_root_credentials(self):
...
+
+    param_list = ['account_name', 'account_id', 'canonical_id', 'mail', 's3_user_id', 'const_cipher_secret_str', 'const_cipher_access_str']
```
vimalraghwani20, line 626: see if we can avoid this list. 
we already have key name in the config file.. 
use the same key name as is...

pranavPawar07: @nileshgovande views??
If we have to again read the keys from the config file, we will again have to either call get_all_keys() which will return all the keys from the config file and parse the list or we will have to create a list of parameters by calling the confstore to get the key name for no. of keys.

s3backgrounddelete/s3backgrounddelete/config/s3_background_delete_config.yaml.sample
```diff
@@ -67,3 +67,13 @@ leakconfig:                                  # Section for object leak config
    leak_processing_delay_in_mins: 15                          # Time delay in mins after which probable delete entry is procesed for leak.
    version_processing_delay_in_mins: 5                        # Time delay in mins after which object version is processed
    cleanup_enabled: False                                     # Flag for cleanup during S3 instance non-active.
+
+#DO NOT CHANGE ANY KEY NAME
```
nileshgovande, line 71: Add these parameters to unsafe attributes

nileshgovande (approved):

PR #1024, opened 2021-07-09
[EOS-19716] ST's for chunked upload

SeagateChaDeepak:
http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Component-Test/job/S3Server-PreMerge-Test/1772/

### DIFF
--- a/st/clitests/jclient_spec.py
+++ b/st/clitests/jclient_spec.py
@@ -361,6 +361,121 @@ for i, val in enumerate(pathstyle_values):
 
     JClientTest('Jclient can download 700K file uploaded in chunked mode').get_object("seagatebucket", "700Kfilec").execute_test().command_is_successful().command_created_file("700Kfilec")
 
+    # ************ 8K Unaligned FILE TEST ************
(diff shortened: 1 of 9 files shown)


st/clitests/jclient_spec.py
```diff
@@ -361,6 +361,121 @@ def get_response_elements(response):
...
+    .command_created_file("16KUnalignedfile")
+
+    JClientTest('Jclient can download 16k Unaligned file uploaded in chunked mode') \
+    .get_object("seagatebucket", "16KUnalignedfilec") \
+    .execute_test() \
+    .command_is_successful() \
+    .command_created_file("16KUnalignedfilec")
+
+    # ************ 1M FILE TEST ************
+    JClientTest('Jclient can upload 1M file') \
```
knrajnambiar76, line 434: In case if we are uploading 1M/16k.... files using s3cmd/aws then via jclient not needed. If not then keep it
jclient test with chunk (of different sizes) are important.

knrajnambiar76 (approved):

PR #1041, opened 2021-07-20
avoid the unrelesed lock after the method returns for issue#1037

anson-lo:
Update s3_memory_pool.c. Avoid the unreleased lock pool->lock after the method returns. This patch for issue#1037.


cortx-admin:
Can one of the admins verify this patch?

PR #1069, opened 2021-08-02
Remove duplicate run_ut() from python_coverage.sh

pranavPawar07:
Signed-off-by: Pranav Diliprao Pawar <pranav.pawar@seagate.com>

### DIFF
--- a/scripts/coverage/python_coverage.sh
+++ b/scripts/coverage/python_coverage.sh
@@ -79,7 +79,7 @@ check_report() {
 
 cleanup
 check_prerequisite
-run_ut
+# run_ut
 run_coverage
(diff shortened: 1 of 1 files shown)


scripts/coverage/python_coverage.sh
```diff
@@ -79,7 +79,7 @@ check_report() {
 
 cleanup
 check_prerequisite
-run_ut
+# run_ut
```
sachitanands, line 82: @pranavPawar07  if we are not running UT then how we are collecting coverage report ?
Could you pls check what is the coverage percentage in report in both cases e..g without running UT and with UT run.

PR #1325, opened 2021-09-29
EOS-24480: Custom access key and secret key implementation.

shrimetta:
# Problem Statement
- Allow user to provide custom access key and secret key when creating S3 account.

# Design
- Feature: [Custom Access key and Secret key](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/690356396/F-45K+Custom+Access+key+id+and+Secret+key)

# Coding
   Checklist for Author
-  [x] Coding conventions are followed and code is consistent

# Testing 
  Checklist for Author
- [ ] Unit and System Tests are added
- [x] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [x] Testing was performed with RPM

# Impact Analysis
  Checklist for Author/Reviewer/GateKeeper
- [x] Interface change (if any) are documented
- [x] Side effects on other features (deployment/upgrade)
- [x] Dependencies on other component(s)

# Review Checklist 
  Checklist for Author
- [x] JIRA number/GitHub Issue added to PR
- [x] PR is self reviewed
- [x] Jira and state/status is updated and JIRA is updated with PR link
- [x] Check if the description is clear and explained

# Documentation
  Checklist for Author
- [x] Changes done to WIKI / Confluence page / Quick Start Guide

Signed-off-by: Shri Metta <shri.metta@seagate.com>

### DIFF
--- a/auth/server/src/main/java/com/seagates3/controller/AccountController.java
+++ b/auth/server/src/main/java/com/seagates3/controller/AccountController.java
@@ -151,6 +151,18 @@ public class AccountController extends AbstractController {
 
         account.setEmail(email);
 
+        AccessKey existingAccessKey;
+        try {
+            existingAccessKey = accessKeyDAO.find(accessKey);
+        } catch (DataAccessException ex) {
+            LOGGER.error("Failed to find access key in ldap -" + ex);
+            return accountResponseGenerator.internalServerError();
+        }
+
+        if (existingAccessKey.exists()) {
+            return accountResponseGenerator.accessKeyAlreadyExists();
+        }
+
         try {
             accountDao.save(account);
         } catch (DataAccessException ex) {
@@ -178,16 +190,6 @@ public class AccountController extends AbstractController {
         }
 
         AccessKey rootAccessKey;
-        try {
-            rootAccessKey = accessKeyDAO.find(accessKey);
-        } catch (DataAccessException ex) {
-          LOGGER.error("Failed to find access key in ldap -" + ex);
-            return accountResponseGenerator.internalServerError();
-        }
-
-        if (rootAccessKey.exists()) {
-            return accountResponseGenerator.accessKeyAlreadyExists();
-        }
 
         try {
             if (accessKey == null) {
@@ -308,7 +310,7 @@ public class AccountController extends AbstractController {
         LOGGER.debug("Creating new access key for account: " + name);
         AccessKey rootAccessKey;
         try {
-            rootAccessKey = createRootAccessKey(root);
+            rootAccessKey = AccessKeyService.createAccessKey(root);
         } catch (DataAccessException ex) {
             return accountResponseGenerator.internalServerError();
         }
@@ -341,21 +343,6 @@ public class AccountController extends AbstractController {
 
         userDAO.save(user);
         return user;
-    }
-
-    /*
-     * Create access keys for the root user.
-     */
-    private AccessKey createRootAccessKey(User root) throws DataAccessException {
-        AccessKey accessKey = new AccessKey();
-        accessKey.setUserId(root.getId());
-        accessKey.setId(KeyGenUtil.createUserAccessKeyId(true));
-        accessKey.setSecretKey(KeyGenUtil.generateSecretKey());
-        accessKey.setStatus(AccessKeyStatus.ACTIVE);
-
-        accessKeyDAO.save(accessKey);
-
-        return accessKey;
     }
 
     @Override
--- a/auth/server/src/main/java/com/seagates3/controller/AccountController.java
+++ b/auth/server/src/main/java/com/seagates3/controller/AccountController.java
@@ -32,7 +32,6 @@ import com.seagates3.dao.UserDAO;
 import com.seagates3.dao.ldap.LDAPUtils;
 import com.seagates3.exception.DataAccessException;
 import com.seagates3.model.AccessKey;
-import com.seagates3.model.AccessKey.AccessKeyStatus;
 import com.seagates3.model.Account;
 import com.seagates3.model.Requestor;
 import com.seagates3.model.Role;
--- a/auth/server/src/main/java/com/seagates3/controller/AccountController.java
+++ b/auth/server/src/main/java/com/seagates3/controller/AccountController.java
@@ -41,6 +41,7 @@ import com.seagates3.response.ServerResponse;
 import com.seagates3.response.generator.AccountResponseGenerator;
 import com.seagates3.s3service.S3AccountNotifier;
 import com.seagates3.util.KeyGenUtil;
+import com.seagates3.service.AccessKeyService;
 import com.seagates3.service.GlobalDataStore;
 
 import io.netty.handler.codec.http.HttpResponseStatus;
@@ -96,6 +97,8 @@ public class AccountController extends AbstractController {
     public ServerResponse create() {
         String name = requestBody.get("AccountName");
         String email = requestBody.get("Email");
+        String accessKey = requestBody.get("AccessKey");
+        String secretKey = requestBody.get("SecretKey");
         Account account;
         int accountCount = 0;
         LOGGER.info("Creating account: " + name);
@@ -176,7 +179,23 @@ public class AccountController extends AbstractController {
 
         AccessKey rootAccessKey;
         try {
-            rootAccessKey = createRootAccessKey(root);
+            rootAccessKey = accessKeyDAO.find(accessKey);
+        } catch (DataAccessException ex) {
+          LOGGER.error("Failed to find access key in ldap -" + ex);
+            return accountResponseGenerator.internalServerError();
+        }
+
+        if (rootAccessKey.exists()) {
+            return accountResponseGenerator.accessKeyAlreadyExists();
+        }
+
+        try {
+            if (accessKey == null) {
+                rootAccessKey = AccessKeyService.createAccessKey(root);
+            } else {
+                rootAccessKey = AccessKeyService.createAccessKey(root,
+                                                    accessKey, secretKey);
+            }
         } catch (DataAccessException ex) {
             return accountResponseGenerator.internalServerError();
         }
--- a/auth/server/src/main/java/com/seagates3/controller/AccountController.java
+++ b/auth/server/src/main/java/com/seagates3/controller/AccountController.java
@@ -151,16 +151,18 @@ public class AccountController extends AbstractController {
 
         account.setEmail(email);
 
-        AccessKey existingAccessKey;
-        try {
-            existingAccessKey = accessKeyDAO.find(accessKey);
-        } catch (DataAccessException ex) {
-            LOGGER.error("Failed to find access key in ldap -" + ex);
-            return accountResponseGenerator.internalServerError();
-        }
+        if (accessKey != null) {
+            AccessKey existingAccessKey;
+            try {
+                existingAccessKey = accessKeyDAO.find(accessKey);
+            } catch (DataAccessException ex) {
+                LOGGER.error("Failed to find access key in ldap -" + ex);
+                return accountResponseGenerator.internalServerError();
+            }
 
-        if (existingAccessKey.exists()) {
-            return accountResponseGenerator.accessKeyAlreadyExists();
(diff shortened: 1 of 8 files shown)


auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -176,7 +179,23 @@ public ServerResponse create() {
 
         AccessKey rootAccessKey;
         try {
-            rootAccessKey = createRootAccessKey(root);
+            rootAccessKey = accessKeyDAO.find(accessKey);
```
shalakadharap, line 182: In case if system generated key, You will search LDAP with NULL key here
Make sure it will not fail with exception or put NULL check otherwise

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -176,7 +179,23 @@ public ServerResponse create() {
...
+            return accountResponseGenerator.internalServerError();
+        }
+
+        if (rootAccessKey.exists()) {
+            return accountResponseGenerator.accessKeyAlreadyExists();
+        }
+
+        try {
+            if (accessKey == null) {
+                rootAccessKey = AccessKeyService.createAccessKey(root);
```
shalakadharap, line 194: Looks like you have copied the accekey generation  function to AccessKeyService class
Kindly remove the method from this class and update its references everywhere.
This way we will avoid code duplication.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
```diff
@@ -40,12 +40,30 @@
...
-                requestBody.get("AccountName"));
+        if(S3ParameterValidatorUtil.isValidName(
+                requestBody.get("AccountName"))) {
+            validParamsCount++;
+        }
+
+        String accessKey = requestBody.get("AccessKey");
+        String secretKey = requestBody.get("SecretKey");
+
+        if (accessKey != null || secretKey != null) {
```
shalakadharap, line 57: NUll check is repeating here and in isValidCustomAccesskeyKey method.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
```diff
@@ -40,12 +40,30 @@
...
+        if (S3ParameterValidatorUtil.isValidEmail(
+                requestBody.get("Email"))) {
+            validParamsCount++;
         }
 
-        return S3ParameterValidatorUtil.isValidName(
-                requestBody.get("AccountName"));
+        if(S3ParameterValidatorUtil.isValidName(
+                requestBody.get("AccountName"))) {
+            validParamsCount++;
```
shalakadharap, line 51: Why we are not returning like before. Why we have taken count?
To avoid multiple return statements?

shalakadharap (changes requested):
Kindly work on requested changes

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: I added this method in the service class as this logic is being used at multiple places. Also we planned to update saving secret key logic (encrypt/decrypt). Hence I had started the refactoring. Kindly let me know if you still think that I should remove it from AccessService class.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shrimetta: Valid cases:

1. Both access key and secret key should not be provided (in this case system will generate the keys).
2. Both access key and secret key should be provided.

Invalid cases:

1. Only access key is given
2. Only secret key is given

To support all these cases we require the null checks at both places:

1. if (accessKey != null || secretKey != null) . This will check the Invalid cases.
2. isValidCustomAccessKey(accessKey) && isValidCustomSecretKey(secretKey). This will validate the Valid cases.
3. The null check in isValidCustomAccessKey and isValidCustomSecretKey will actually validate the keys.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shrimetta: I took count for two reasons:

1. Yes to avoid multiple return statements. This makes readability better.
2. Easy to make complex validations. Or it will if () { return true } else { return false}

Also we can update the count as anyways we will be updating the method.

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: Will fix it.

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shalakadharap: I think you misunderstood.
I am telling to remove from AccountController. 
Keep everything under AccessService and update references everywhere.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shalakadharap: This is hardcoding so not looking good here
You can take 1 flag instead of count.
eg- isValidationFailed = False;
As soon as any validation fails, you will set the flag to True.
You will include 1 more condition as follow-
if(!isValidationFailed  && S3ParameterValidatorUtil.isValidCustomAccessKey(accessKey))
and then have a single return statement at the end.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shalakadharap: Why we are combining the 2 checks?
Why not to have separate condition check?
We making it complex

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: Will fix this.

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shrimetta: Will fix as above

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shalakadharap: Keep null check at both. As discussed implement rest

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: Done

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: Done

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shrimetta: Done

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shrimetta: Done

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shalakadharap: cant see this fixed

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
```diff
@@ -40,12 +40,27 @@
...
+                                                requestBody.get("AccountName"));
         }
 
-        return S3ParameterValidatorUtil.isValidName(
-                requestBody.get("AccountName"));
+        if (isValidParameters) {
+            String accessKey = requestBody.get("AccessKey");
+            String secretKey = requestBody.get("SecretKey");
+
+            if (accessKey != null || secretKey != null) {
```
shalakadharap, line 55: This condition has to be &&

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: null check was missed in the commit. Added it.
Done

auth/server/src/main/java/com/seagates3/parameter/validator/AccountParameterValidator.java
shrimetta: If one of the key is not null and another is null, we still want to validate and send error to user that he has only provided one parameter and he should provide both. If we use && then it will not validate and isValidParameters will remain true.

bkirunge7 (commented):
Please add UT and ST's

auth/server/src/main/java/com/seagates3/controller/AccountController.java
```diff
@@ -148,6 +151,20 @@ public ServerResponse create() {
 
         account.setEmail(email);
 
+        if (accessKey != null) {
+            AccessKey existingAccessKey;
+            try {
+                existingAccessKey = accessKeyDAO.find(accessKey);
+            } catch (DataAccessException ex) {
+                LOGGER.error("Failed to find access key in ldap -" + ex);
```
bkirunge7, line 159: Is this error is about failing to read from LDAP? and not about access key is not found in LDAP

auth/server/src/main/java/com/seagates3/service/AccessKeyService.java
```diff
@@ -52,5 +53,41 @@ public static AccessKey createFedAccessKey(User user, long timeToExpire)
...
+            throws DataAccessException {
+        AccessKeyDAO accessKeyDAO = (AccessKeyDAO) DAODispatcher.getResourceDAO(
+                DAOResource.ACCESS_KEY);
+        AccessKey accessKey = new AccessKey();
+        accessKey.setUserId(user.getId());
+        accessKey.setId(KeyGenUtil.createUserAccessKeyId(true));
+        accessKey.setSecretKey(KeyGenUtil.generateSecretKey());
+        accessKey.setStatus(AccessKeyStatus.ACTIVE);
+
+        accessKeyDAO.save(accessKey);
```
bkirunge7, line 70: call AccessKey createAccessKey(User user, String accessKeyId,
            String secretKey) throws DataAccessException {

to save access key. This function can just generate access key

shalakadharap (approved):
Make sure to Test all scenarios and add UTs

auth/server/src/main/java/com/seagates3/controller/AccountController.java
shrimetta: This error is about failing to read from LDAP.

auth/server/src/main/java/com/seagates3/service/AccessKeyService.java
shrimetta: Done

bkirunge7 (approved):

nileshgovande:
@shrimetta - Pls pass s3 mini provisioning job on this branch
Also what testing is done on it ?

nileshgovande:
Link - http://eos-jenkins.colo.seagate.com/job/Cortx-PR-Build/job/Mini-Provisioning/job/CentOS-7.9.2009/job/S3-Server/

nileshgovande:
In my mind - these changes should go to main and they have to come to tk8s through a rebase

kupranay:
@nileshgovande  this change is a requirement for LC and as of today not a requirement for LR. As per our initial discussion, we should minimize changes in on main and put there which is coming from past. All the specific requirement for LC should go to Kubernatese branch. Hence this changes should be in kubernatese branch as per me

shrimetta:
Closing this PR. Will create a fresh PR with all review comments implemented.

